PR #5, opened 2019-02-21
Fix a bunch of warnings

vanwagonet:
There should not be any semantic changes here, just trying to cut down the number of warnings.

refs: none
affects: none
release note: none


CanvasCore/CanvasCore/Helm/Extensions/HelmSplitViewControllerExtensions.swift
```diff
@@ -16,7 +16,7 @@
 
 import UIKit
 
-public extension HelmSplitViewController {
+extension HelmSplitViewController {
```
vanwagonet: This `public` causes the `description`'s `open` to be a warning, but changing that to `public` is an error.

CanvasCore/CanvasCore/Utilities/NSError+SoLazy.swift
```diff
@@ -89,7 +89,7 @@ extension NSError {
         var report = "===== Error Report \(domain)–\(code) =====\n"
         
         for (key, value) in userInfo {
-            if (key as? String) == NSUnderlyingErrorKey { continue } // handled separately
+            if key == NSUnderlyingErrorKey { continue } // handled separately
```
vanwagonet: already a `String`

CanvasCore/CanvasCore/WhizzyWig/CanvasWebView.swift
```diff
@@ -200,8 +200,7 @@ public class CanvasWebView: WKWebView {
     
     @objc public func htmlContentHeight(completionHandler: @escaping (CGFloat) -> Void) {
         evaluateJavaScript("document.documentElement.scrollHeight") { result, error in
-            guard let height = result as? NSNumber else { completionHandler(0.0); return }
-            completionHandler(CGFloat(height))
```
vanwagonet: This initializer is deprecated, just cast straight to `CGFloat` instead of through `NSNumber`.

Parent/Parent/Browser/WebBrowserViewController.swift
```diff
@@ -153,7 +153,7 @@ class WebBrowserViewController: UIViewController {
 
     @objc func reloadButtonTapped(_ refreshButton: UIBarButtonItem) {
         networkOps = 0
-        if let request = webView.request, (request.url?.absoluteString.characters.count)! > 0 && request.url?.absoluteString != "about:blank" {
+        if let request = webView.request, request.url?.absoluteString.isEmpty == false && request.url?.absoluteString != "about:blank" {
```
vanwagonet: `.characters` is deprecated. It was silly to use it here anyway.

Canvas/CanvasKit1/UI Classes/CKCommentViewController.m
```diff
@@ -326,7 +326,7 @@ - (IBAction)flipInputPanel:(id)sender
         } completion:^(BOOL finished){
             [UIView animateWithDuration:0.2
                              animations:^{
-                                 [self.recorderView setMode:preferredMediaCommentMode];
+                                 [self.recorderView setMode:self->preferredMediaCommentMode];
```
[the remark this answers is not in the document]
armstrongnate: `self.` is safer. `self->` will not trigger the getter.

CanvasCore/CanvasCore/Helm/Extensions/HelmSplitViewControllerExtensions.swift
armstrongnate: Nice. Yeah `public extension` isn't a thing.

Parent/Parent/Browser/WebBrowserViewController.swift
armstrongnate: Better in so many ways.

vanwagonet:
I was able to switch a few to `self.`, but most of them are not properties and give a compiler error unless you use `self->`.

PR #60, opened 2019-03-15
Show submission files in comments

vanwagonet:
refs: MBL-12160
affects: none
release note: none


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  One or more files in Core are below the minimum test coverage 50%
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  One or more files in Student are below the minimum test coverage 50%
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: None

[MBL-12160](https://instructure.atlassian.net/browse/MBL-12160)

**Core** | 91.38% | 91.41% | -0.03%
Core/Core/Profile/DrawerTransitioning.swift | 0% | 0% | 0%
Core/Core/Router/Route.swift | 42.86% | 42.86% | 0%
**Student** | 91.18% | 94.68% | -3.5%
Student/Student/Submissions/SubmissionComments/SubmissionCommentFileView.swift | 0% | -- | --
<p align="right">
</p>

Core/Core/Database/CoreData/Models/Submission.swift
```diff
@@ -145,11 +146,64 @@ extension Submission: WriteableModel {
             model.mediaComment = try MediaComment.save(mediaComment, in: client)
         }
 
-        SubmissionComment.save(item, in: client)
+        if item.submission_type != nil, item.submission_type != .none, item.submission_type != .not_graded, item.submission_type != .on_paper {
+            SubmissionComment.save(item, in: client)
```
bluwave: I'm not sure of the answer here so just posing this question, I didn't think we do a lot of `EntityName.save` because I thought most db operations were wrapped in some sort of persistence background thread block or db operation that then calls save at the end.  Do we really need this here?  Seems to follow it's own pattern versus the current pattern of the code as a whole.

Core/Core/Database/CoreData/Models/Submission.swift
```diff
@@ -145,11 +146,64 @@ extension Submission: WriteableModel {
...
+        }
         for comment in item.submission_comments ?? [] {
             SubmissionComment.save(comment, forSubmission: item.id.value, in: client)
```
bluwave: same here.  ( I know you didn't change this line as part of the fix for this)

Core/Core/Extensions/UIViewExtensions.swift
```diff
@@ -16,6 +16,19 @@
 
 import UIKit
 
+public protocol ViewLoader {}
+extension UIView: ViewLoader {}
+extension ViewLoader where Self: UIView {
+    /// Returns a newly initialized view controller.
+    /// This can assume the nib name matches the type name and the bundle contains the type.
+    public static func loadFromXib(nibName name: String = String(describing: Self.self)) -> Self {
```
bluwave: nice

Core/Core/Database/CoreData/Models/Submission.swift
shichongrui: I think these methods are different than the overall `save` method that is called at the end of the db operation. These could be called something different than `save` to disambiguate them from the database client `save` method. These are more so `construct` methods or something to that affect.

shichongrui: But I think the idea was that we want the models to know how to construct themselves rather than a use case or a parent model having to know how to construct their dependencies.

vanwagonet: I think this is the pattern of the code as a whole. A use case's write calls `Model.save`. When that model has sub-models, it recursively calls `Model.save` for those.

The problem here was that we created a comment for submission types that shouldn't show a comment. Rather than trying to filter it out later, we just don't make a comment entry for those submissions.

vanwagonet: I guess there is something unique here, since we create a `SubmissionComment` from an `APISubmission`. That is because we generally show each submission attempt in the comments view mixed in with the actual comments.

Student/Student/Submissions/SubmissionComments/SubmissionCommentAttemptCell.swift
```diff
@@ -0,0 +1,53 @@
...
+    func update(comment: SubmissionComment, submission: Submission?, onFileTap: @escaping (Submission?, File?) -> Void) {
+        accessibilityIdentifier = "SubmissionCommentsElement.attemptCell.\(comment.id)"
+        accessibilityLabel = String.localizedStringWithFormat(
+            NSLocalizedString("On %@ %@ submitted the following", bundle: .student, comment: ""),
+            DateFormatter.localizedString(from: comment.createdAt, dateStyle: .long, timeStyle: .short),
+            comment.authorName
+        )
+        self.onFileTap = onFileTap
+
+        for view in stackView?.arrangedSubviews ?? [] { view.removeFromSuperview() }
```
bluwave: At first glance I'm wondering if any of this is affecting scroll performance?  Is this cell scrolling choppy at all?  It may not be but something we may want to keep an eye on here.

vanwagonet: Maybe a little choppy here and there? If you know of a better way to dynamically replace a list of views in a reused cell, I'd love to make this smoother.

Core/Core/Database/CoreData/Models/Submission.swift
bluwave: Sounds good.  Thanks for helping me understand this part.

PR #66, opened 2019-03-19
Add support for masquerading in Parent

shichongrui:
I found that the Not a Parent view was still in React Native and needs to be refactored. Will do it in a subsequent commit.

refs: MBL-12161
affects: Parent
release note: none
test plan:
 - All options in the profile should work
 - Masquerading should work
 - Tapping the version number 10 times should still show the dev menu
 - Developer menu should still work


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please run `yarn update-headers`. The following do not have the correct license header: 
* Core/CoreTests/API/APIPermissionsRequestableTests.swift
* Core/CoreTests/Database/Models/PermissionsTests.swift
* Core/CoreTests/Use Cases/GetContextPermissionsTests.swift
* TestsFoundation/TestsFoundation/Fixtures/Api/APIPermissions.swift
* TestsFoundation/TestsFoundation/Fixtures/Model/PermissionsFixture.swift
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Parent

[MBL-12161](https://instructure.atlassian.net/browse/MBL-12161)

**Core** | 94.26% | 94.16% | 0.1%
**Student** | 94.83% | 94.83% | 0%
<p align="right">
</p>

Core/Core/ActAsUser/ActAsUserViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+        userIDTextField.placeholder = NSLocalizedString("User ID", bundle: .core, comment: "")
+        actAsUserButton.titleLabel?.text = NSLocalizedString("Act as User", bundle: .core, comment: "")
+        domainTextField.text = env?.currentSession?.baseURL.absoluteString
+
+        domainTextField.addTarget(self, action: #selector(updateActAsUserButtonDisabledStatus), for: .editingChanged)
+        userIDTextField.addTarget(self, action: #selector(updateActAsUserButtonDisabledStatus), for: .editingChanged)
+        actAsUserButton.isEnabled = false
+
+        domainTextField.addTarget(self, action: #selector(enterPressed), for: .editingDidEndOnExit)
+        userIDTextField.addTarget(self, action: #selector(enterPressed), for: .editingDidEndOnExit)
```
vanwagonet: I believe a bunch of these targets can be set up in the storyboard or xib. Setting them up here is fine though.

Core/Core/ActAsUser/ActAsUserViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+
+        redPanda.transform = redPanda.transform.rotated(by: .pi / -6).translatedBy(x: 0, y: 0)
+        actAsUserButton.roundCorners(corners: [.topLeft, .topRight, .bottomLeft, .bottomRight], radius: 5)
```
vanwagonet: if you are rounding all of the corners `layer.cornerRadius = 5` is probably simpler. I'm not sure about performance difference. That can also be added as a custom property in the storyboard.

Core/Core/ActAsUser/ActAsUserViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+                self?.showMasqueradingError()
+            }
+        }
+    }
+}
+
+extension ActAsUserViewController {
+    @objc
+    func updateActAsUserButtonDisabledStatus() {
+        self.actAsUserButton.isEnabled = domainTextField.text?.count ?? 0 > 0 && userIDTextField.text?.count ?? 0 > 0
```
vanwagonet: `.text?.isEmpty == false` No need to count the characters, since we only care if it's empty or not.

Core/Core/Database/CoreData/Models/Permissions.swift
```diff
@@ -0,0 +1,173 @@
...
+    @NSManaged public var viewAuditTrail: Bool
+    @NSManaged public var viewGroupPages: Bool
+    @NSManaged public var viewUserLogins: Bool
+
+    @discardableResult
+    public static func save(_ item: APIPermissions, for context: Context, in client: PersistenceClient) throws -> Permissions {
+        let predicate = NSPredicate(format: "%K == %@", #keyPath(Permissions.context), context.canvasContextID)
+        let model: Permissions = client.fetch(predicate).first ?? client.insert()
+        model.context = context.canvasContextID
+        model.becomeUser = item.become_user ?? model.becomeUser
```
vanwagonet: The way we request these, we only want to change the ones we got back, right? If it doesn't come back, we can't assume it's `false`. That's what this code looks like it's doing. Did I understand right?

Core/Core/Use Cases/GetContextPermissions.swift
```diff
@@ -0,0 +1,51 @@
...
+    let context: Context
+    let permissions: [PermissionName]
+
+    public init(context: Context, permissions: [PermissionName] = []) {
+        self.context = context
+        self.permissions = permissions
+    }
+
+    public var cacheKey: String? {
+        return "get-\(context.canvasContextID)-permissions-\(permissions.map({ $0.rawValue }).joined(separator: ","))"
```
vanwagonet: maybe a `sort()` in there for good measure?

Core/Core/Use Cases/GetContextPermissions.swift
```diff
@@ -0,0 +1,51 @@
...
+            order: []
+        )
+    }
+
+    public func write(response: APIPermissions?, urlResponse: URLResponse?, to client: PersistenceClient) throws {
+        guard let item = response else {
+            return
+        }
+        try Permissions.save(item, for: context, in: client)
+    }
```
vanwagonet: `Permissions` looks like it could be marked a `WriteableModel`, and then we'd not need a custom `write` here.

Parent/Parent/Profile/ProfilePresenter.swift
```diff
@@ -19,52 +19,111 @@ import Core
...
+        }
+
+        if env.api.baseURL.absoluteString.contains("siteadmin.instructure.com") {
+            return true
+        }
+
+        guard let becomeUser = self.permissions.first?.becomeUser else {
+            return false
+        }
+        return becomeUser
```
vanwagonet: `return permissions.first?.becomeUser ?? false`

Parent/Parent/Profile/ProfilePresenter.swift
```diff
@@ -19,52 +19,111 @@ import Core
...
+    }
+
     public var cells: [ProfileViewCell] {
-        return [
+        var cells: [ProfileViewCell] = []
+
+        if self.canMasquerade {
```
vanwagonet: I like these conditions much better than the `hidden` flag from before. Good improvement.

scripts/coverage/config.json
```diff
@@ -9,6 +9,7 @@
     "Core\/ShareExtension\/",
     "Core\/Use Cases\/SubmitFileSubmission",
     "OperationQueueExtensions.swift",
+    "DrawerTransitioning.swift",
```
vanwagonet: ` "Transitioning.\\w+$"` was already added above.

AllTheThings.xcworkspace/xcshareddata/IDETemplateMacros.plist
```diff
@@ -6,17 +6,17 @@
...
-// GNU General Public License for more details.
+// http://www.apache.org/licenses/LICENSE-2.0
 //
-// You should have received a copy of the GNU General Public License
-// along with this program.  If not, see &lt;http://www.gnu.org/licenses/&gt;.
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
```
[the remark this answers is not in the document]
shichongrui: Danger was failing telling me that license headers on new files were incorrect. So I ran the update script and took the license header that it put in the files and put it in here. Is it not correct or do not all new files get this header?

Core/Core/ActAsUser/ActAsUserViewController.swift
shichongrui: I was trying to do `cornerRadius` but it was saying it was not a property on a `UIButton`. Didn't realize I had to do it on the `layer`. I will make that change.

Core/Core/Database/CoreData/Models/Permissions.swift
shichongrui: Correct. If we previously get all of the permissions for the user but then only request one permission later, we don't want to overwrite all of the existing permissions. If we have no permissions and only request one, we want to set the rest as false until we know from a subsequent request that they are true.

Core/Core/Use Cases/GetContextPermissions.swift
shichongrui: Good thinking. I will add that.

Core/Core/Database/CoreData/Models/Permissions.swift
shichongrui: I think I misread your comment the first time in regards to assuming the permission is false. I initially went the route of marking the permissions as `Bool?` but that is not a possible type for an `@NSManaged` property. So I either had to use `NSBool` or make them false by default. I think it's safe to mark them as false by default. I'd rather that than having them be true by default. I would expect a screen that needs to know about a permission to request its true value on load. Does that seem reasonable?

Core/Core/ActAsUser/ActAsUserViewController.swift
shichongrui: Oh great suggestion.

AllTheThings.xcworkspace/xcshareddata/IDETemplateMacros.plist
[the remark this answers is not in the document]
shichongrui: I reverted this change. I saw in the update-header script that it depends on the file type.

Core/Core/Use Cases/GetContextPermissions.swift
vanwagonet: oh, right. good call.

Core/Core/Database/CoreData/Models/Permissions.swift
vanwagonet: Yeah, I think that's reasonable.

Core/Core/Profile/ProfileViewController.swift
```diff
@@ -19,24 +19,29 @@ import UIKit
...
+    var view: ProfileViewControllerProtocol? { get set }
     var cells: [ProfileViewCell] { get }
     func didTapVersion()
+    func viewIsReady()
 }
 
+public protocol ProfileViewControllerProtocol: class {
+    func reload()
+    func show(_ route: Route, options: Core.Router.RouteOptions?)
+    func show(_ route: String, options: Core.Router.RouteOptions?)
```
vanwagonet: Hmm, I would have thought routing should be done in the presenter rather than the view controller protocol.

Parent/Parent/ParentAppDelegate.swift
```diff
@@ -275,7 +275,14 @@ extension ParentAppDelegate: NativeLoginManagerDelegate {
         guard let token = client.authSession.token else {
             return
         }
-        let entry = KeychainEntry(accessToken: token, baseURL: client.authSession.baseURL, expiresAt: nil, locale: "en", masquerader: nil, refreshToken: nil, userAvatarURL: client.authSession.user.avatarURL, userID: client.authSession.user.id, userName: client.authSession.user.name)
+
+        let masquerader = client.originalBaseURL != nil
```
vanwagonet: I think `originalBaseURL` might be nil if you masquerade within your same org.

Core/Core/Profile/ProfileViewController.swift
shichongrui: The profile view controller has to dismiss itself prior to the presenter being able to route to anything. Because we  don't want a hard dependency on the view controller, the protocol must either define the dismiss method or, as it was done in bridge, the view controller has defined `show` methods that takes care of dismissing itself and routing.

Parent/Parent/ParentAppDelegate.swift
shichongrui: Oh good to know. I'll change this up.

Core/Core/Profile/ProfileViewController.swift
vanwagonet: ok, I think that makes sense, that the profile is an exception to the usual pattern.

shichongrui: Yeah kind of sucks but someway we have to expose some way to the presenter to handle the dismiss.

PR #80, opened 2019-03-28
Add ModuleItem

armstrongnate:
[ignore-commit-lint]


inst-danger:
**Core** | 94.05% | 94.03% | 0.01%
**Student** | 93.01% | 93.01% | 0%
**React Native** | 92.53% | 92.53% | 0%
<p align="right">
</p>

Core/Core/Database/CoreData/Models/Module.swift
```diff
@@ -23,6 +23,7 @@ public class Module: NSManagedObject {
     @NSManaged public var position: Int
     @NSManaged public var courseID: String
     @NSManaged public var published: Bool
+    @NSManaged public var items: Set<ModuleItem>?
```
vanwagonet: aren't the items supposed to be ordered?

Core/Core/Database/CoreData/Models/ModuleItem.swift
```diff
@@ -0,0 +1,65 @@
...
+    @NSManaged public var title: String
+    @NSManaged public var indent: Int
+    @NSManaged public var htmlURL: URL
+    @NSManaged public var url: URL?
+    @NSManaged public var published: Bool
+    @NSManaged public var module: Module?
+    @NSManaged public var primitiveType: Data?
+    public var type: ModuleItemType? {
+        get {
+            willAccessValue(forKey: "type")
```
vanwagonet: why do we need the explicit will & did?

armstrongnate: They are needed to trigger change notifications and [setting primitive values will not call them](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreData/LifeofaManagedObject.html).

I was thinking `primitiveType` is want I want here but we've been using `fooRaw` everywhere else. I guess I could change this to `typeRaw` and have a custom getter and setter through `type` and not bother with `willAccessValue` and `didAccessValue`.

Core/Core/Database/CoreData/Models/Module.swift
armstrongnate: I was thinking I would order them via the NSFetchedResultsController. I don't think they need to be ordered here.

Would I just use a `NSOrderedSet`?

Core/Core/Database/CoreData/Models/ModuleItem.swift
armstrongnate: I cleaned this up with `typeRaw` instead of `primitiveType`.

Core/Core/Database/CoreData/Models/Module.swift
vanwagonet: if we aren't going to access them from this, why create the relationship?

armstrongnate: Good point. I added it for a test. I will remove the relationship.

armstrongnate: Replaced relationship with `courseID` reference.

PR #85, opened 2019-04-01
Fix installing Eureka pod

armstrongnate:
refs: none
affects: parent
release note: none

I couldn't find a version of Eureka that builds with Swift 3 so we're
pulling in a Swift 4.2 version. I had to tweak the Podfile in order to
build Eureka with Swift 4.2.

### DIFF
--- a/Parent/Parent/Thresholds/StudentSettingsViewController.swift
+++ b/Parent/Parent/Thresholds/StudentSettingsViewController.swift
@@ -102,7 +102,7 @@ open class StudentSettingsViewController : FormViewController {
         let sectionTitle = NSLocalizedString("Alert me when:", comment: "Alert Section Header")
         form +++
             Section() {
-                $0.header = studentSectionHeaderView()
+                $0.header = self.studentSectionHeaderView()
             }
             +++ Section(sectionTitle)
             <<< rowForThresholdType(.courseGradeHigh)
--- a/Podfile
+++ b/Podfile
@@ -67,7 +67,7 @@ abstract_target 'defaults' do
   target 'Parent' do
     project 'Parent/Parent.xcodeproj'
     pod 'Fabric', '~> 1.7.7'
-    pod 'Eureka', :git => 'https://github.com/xmartlabs/Eureka', :branch => 'Swift-3.3'
+    pod 'Eureka', '~> 4.3'
     pod 'Firebase/Core'
   end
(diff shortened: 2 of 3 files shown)


Podfile
```diff
@@ -125,8 +125,11 @@ end
 
 post_install do |installer|
   installer.pods_project.targets.each do |target|
+    usePackageSwift = %w[ Eureka ]
     target.build_configurations.each do |config|
-      config.build_settings['SWIFT_VERSION'] = '3.0'
+      unless usePackageSwift.include? target.name
+        config.build_settings['SWIFT_VERSION'] = '3.0'
+      end
```
vanwagonet: I like this approach. I figured we were going to need to do this as we migrated anyway, so it's good to see it look nice in practice.

Parent/Parent/Thresholds/StudentSettingsViewController.swift
```diff
@@ -102,7 +102,7 @@ open class StudentSettingsViewController : FormViewController {
         let sectionTitle = NSLocalizedString("Alert me when:", comment: "Alert Section Header")
         form +++
             Section() {
-                $0.header = studentSectionHeaderView()
+                $0.header = self.studentSectionHeaderView()
```
bluwave: should this be weak or should no worries here?

armstrongnate: I think we are good. `Section` does not hold a reference to the `initializer` block.

This is what it looks like:

```swift
    public init(_ initializer: @escaping (Section) -> Void) {
        initializer(self)
    }
```

Not really sure why they mark is as `@escaping`.

armstrongnate: Also, we don't hold a reference to the Section. Looks like it's just used to build a view then de-initialize.

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Parent

**Core** | 94.51% | 94.51% | 0%
**Student** | 93.29% | 93.29% | 0%
**React Native** | 92.53% | 92.53% | 0%
<p align="right">
</p>

PR #97, opened 2019-04-04
Fix ModuleListPresenter tests

armstrongnate:
[ignore-commit-lint]

The main issue here was that `UseCase.fetch` callbacks were being
completed in tests that did not initiate them.

- TestA kicks off a request for courses and a request for modules
- TestA waits until courses load then the test would finish
- TestB starts
- TestB mocks modules request
- TestB requests modules
- TestA's request for modules would finish
- **TestA's modules would get written to TestB's db**
- TestB's request for modules would finish
- TestB's modules would get written to TestB's db
- TestB would fail because too many modules were written

This is due to UseCases retaining themselves until the callbacks
complete and not having a way to cancel a UseCase..

We should _probably_ support canceling at some point but in the
meantime, ensuring that we write to the same database that initiated the
`fetch` fixes test issues since we re-create the db for every test.

### DIFF
--- a/Core/Core/Extensions/DispatchExtensions.swift
+++ b/Core/Core/Extensions/DispatchExtensions.swift
@@ -1,0 +1,30 @@
+//
+// Copyright (C) 2019-present Instructure, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, version 3 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+import Foundation
+
+/// Performs `closure` on the main thread
+///
+/// If the executing thread is already the main thread, the `closure`
+/// is executed immediately.
+public func performUIUpdate(using closure: @escaping () -> Void) {
+    // If we are already on the main thread, execute the closure directly
+    if Thread.isMainThread {
+        closure()
+    } else {
+        DispatchQueue.main.async(execute: closure)
+    }
+}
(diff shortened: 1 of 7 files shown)


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please run `yarn update-headers`. The following do not have the correct license header: 
* Core/CoreTests/Extensions/DispatchExtensionsTests.swift
  </td>
    </tr>
  </tbody>
</table>



**Core** | 94.13% | 94.14% | -0%
**Student** | 93.36% | 93.36% | 0%
<p align="right">
</p>

Core/Core/Extensions/DispatchExtensions.swift
```diff
@@ -0,0 +1,30 @@
...
+/// Performs `closure` on the main thread
+///
+/// If the executing thread is already the main thread, the `closure`
+/// is executed immediately.
+public func performUIUpdate(using closure: @escaping () -> Void) {
+    // If we are already on the main thread, execute the closure directly
+    if Thread.isMainThread {
+        closure()
+    } else {
+        DispatchQueue.main.async(execute: closure)
```
bluwave: clever

Core/Core/Store/UseCase.swift
```diff
@@ -63,6 +63,9 @@ extension UseCase {
     }
 
     public func fetch(environment: AppEnvironment = .shared, force: Bool = false, _ callback: @escaping RequestCallback) {
+        // Make sure we write to the database that initiated this request
+        let database = environment.database
```
bluwave: i'm lost here, was another environment being used than the one passed in?

bluwave:
would the `tearDown` method in `CoreTestCase` be of any help to reset useCases, callbacks, or database?

shichongrui:
Looks like there are still some flaky tests.

Core/Core/Store/UseCase.swift
armstrongnate: The `environment` doesn't change but `environment.database` does change between the time the `fetch` starts and completes.

armstrongnate:
@bluwave Ideally we would cancel the use cases on tearDown but we don't have a way to do that right now.

@shichongrui Thanks, I'll look at the flaky test.

I'm glad this pr is exposing a bunch of flakiness.

PR #147, opened 2019-04-30
Enable students to edit pages

shichongrui:
refs: MBL-12293
affects: Student
release note: Made it possible for students to edit pages
test plan: see ticket


inst-danger:
#### Release Note: 
Made it possible for students to edit pages

#### Affected Apps: Student

[MBL-12293](https://instructure.atlassian.net/browse/MBL-12293)

**Core** | 93.6% | 93.62% | -0.01%
**Student** | 92.2% | 92.2% | 0%
**React Native** | 92.65% | 92.63% | 0.02%
<p align="right">
</p>

rn/Teacher/src/modules/pages/details/PageDetails.js
```diff
@@ -57,7 +57,7 @@ export class PageDetails extends Component<Props> {
   constructor (props: Props) {
     super(props)
 
-    if (isStudent) {
+    if (isStudent()) {
```
armstrongnate: Oh boy. That's awesome.

rn/Teacher/src/modules/pages/edit/PageEdit.js
```diff
@@ -142,51 +147,54 @@ export class PageEdit extends Component<Props, State> {
...
-                onValueChange={this.handleEditingRolesChange}
-                testID='pages.edit.editing_roles.picker'
-              >
-                {Object.keys(editingRoles()).map(key => (
-                  <PickerItem
-                    key={key}
-                    value={key}
-                    label={editingRoles()[key]}
+            {isTeacher() &&
+              <React.Fragment>
```
armstrongnate: Interesting. What is `React.Fragment`?

rn/Teacher/src/modules/inbox/__tests__/__snapshots__/Compose.test.js.snap
```diff
@@ -284,15 +284,96 @@ exports[`Compose renders after picking a course 1`] = `
...
                 },
-                undefined,
+                Object {
+                  "color": "#2D3B45",
+                },
               ]
             }
           >
-            Select a Course
+            Learn React Native
```
armstrongnate: The changes in this file seem odd. Do you know what they are about?

rn/Teacher/src/modules/pages/edit/PageEdit.js
shichongrui: `React.Fragment` is a component that essentially just wraps other components. So in this case when we want to conditionally render a couple of views that are all siblings, without `React.Fragment` we have 3 options.
 - We conditionally render each sibling individually which isn't great. 
 - We wrap them in a `View` and conditionally render that, but then we have an extra `View` hanging around that isn't really needed
 - We conditionally render an array of views. But with this then you have to provide a key for each element which isn't great.

So instead we use `React.Fragment` which is a wrapping component that renders nothing. Gives us the ability to conditionally render a bunch of sibling components without the extra `View`.

rn/Teacher/src/modules/inbox/__tests__/__snapshots__/Compose.test.js.snap
shichongrui: These changes happened after I updated enzyme to work with `React.Fragment`. I think these changes are actually correct and that what was there before was not. The test is testing something after choosing a course, but the original snapshot still said `Select a Course`. So it seemed like the enzyme update actually made the test work the way it was intended.

PR #187, opened 2019-05-15
Fix crash when course is restricted by date

armstrongnate:
refs: MBL-12311
affects: student
release note: Fixed a crash in Inbox when a student was enrolled in a past or future course

Fixes the crash on the course filter as well as fixes a related bug in the
`Compose` course picker.

### DIFF
--- a/rn/Teacher/src/modules/app/index.js
+++ b/rn/Teacher/src/modules/app/index.js
@@ -25,6 +25,7 @@ export type App = {
 const teacher = {
   appId: 'teacher',
   filterCourse: (course: Course): boolean => {
+    if (course.access_restricted_by_date) return false
     const enrollments = course.enrollments
     if (!enrollments) return false
     return enrollments.some((e) =>
@@ -40,7 +41,7 @@ const teacher = {
 
 const student = {
   appId: 'student',
-  filterCourse: (course: Course): boolean => true,
+  filterCourse: (course: Course): boolean => !course.access_restricted_by_date,
 }
 
 const parent = {
--- a/rn/Teacher/src/modules/inbox/CourseSelect.js
+++ b/rn/Teacher/src/modules/inbox/CourseSelect.js
@@ -110,7 +110,6 @@ export function mapStateToProps (state: AppState): CourseSelectDataProps {
   let courses = Object.keys(state.entities.courses)
     .map(id => state.entities.courses[id].course)
     .filter(App.current().filterCourse)
-    .filter(course => !course.access_restricted_by_date)
   let pending = !!state.favoriteCourses.pending
 
   const favoriteCourses = courses.filter((course) => course.is_favorite)
--- a/rn/Teacher/src/modules/inbox/CourseSelect.js
+++ b/rn/Teacher/src/modules/inbox/CourseSelect.js
@@ -110,6 +110,7 @@ export function mapStateToProps (state: AppState): CourseSelectDataProps {
   let courses = Object.keys(state.entities.courses)
     .map(id => state.entities.courses[id].course)
     .filter(App.current().filterCourse)
+    .filter(course => !course.access_restricted_by_date)
   let pending = !!state.favoriteCourses.pending
 
   const favoriteCourses = courses.filter((course) => course.is_favorite)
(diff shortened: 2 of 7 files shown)


rn/Teacher/src/modules/inbox/CourseSelect.js
```diff
@@ -110,6 +110,7 @@ export function mapStateToProps (state: AppState): CourseSelectDataProps {
   let courses = Object.keys(state.entities.courses)
     .map(id => state.entities.courses[id].course)
     .filter(App.current().filterCourse)
+    .filter(course => !course.access_restricted_by_date)
```
armstrongnate: I don't know if I should just add this to `App.current().filterCourse`. The Dashboard makes heavy use of it and I don't know if there's ever a case where we'd show a course that has `access_restricted_by_date = true` on the Dashboard, all courses, invites, etc. Any thoughts?

rn/Teacher/src/modules/inbox/__tests__/Inbox.test.js
```diff
@@ -53,258 +54,302 @@ let defaultProps = {
...
-
-  props = {
-    refreshInboxArchived: jest.fn(),
-    scope: 'archived',
-    refreshCourses: jest.fn(),
-  }
-  handleRefresh(props)
-  expect(props.refreshInboxArchived).toHaveBeenCalled()
-})
+  it('shows CourseFilter', () => {
```
armstrongnate: This is one of the two new tests I added. Everything else was moved to be nested within `describe('Inbox', () => {`

rn/Teacher/src/modules/inbox/__tests__/Inbox.test.js
```diff
@@ -53,258 +54,302 @@ let defaultProps = {
...
-  const props = {
-    refreshInboxArchived: jest.fn(),
-    scope: 'archived',
-    refreshCourses: jest.fn(),
-    courses: [template.course()],
-  }
-  handleRefresh(props)
-  expect(props.refreshCourses).not.toHaveBeenCalled()
-})
+  it('hides CourseFilter when no courses', () => {
```
armstrongnate: This is the other test I added.

rn/Teacher/src/modules/inbox/__tests__/Inbox.test.js
```diff
@@ -53,258 +54,302 @@ let defaultProps = {
...
 
-  it('returns false when there is next & conversations', () => {
-    expect(shouldRefresh({
-      conversations: [ template.conversation({}) ],
-      scope: 'all',
-      courses: [],
-      next () {},
-      navigator: new Navigator(''),
-    })).toBe(false)
+  it('filters courses restricted by date', () => {
```
armstrongnate: I also added this test.

rn/Teacher/src/modules/inbox/CourseSelect.js
vanwagonet: Yeah, honestly I'd rather they not come back in the api at all. As far as I know, we never show a restricted course. The earlier we can filter them out the better.

inst-danger:
#### Release Note: 
Fixed a crash in Inbox when a student was enrolled in a past or future course

#### Affected Apps: Student

[MBL-12311](https://instructure.atlassian.net/browse/MBL-12311)

**Core** | 93.55% | 93.55% | 0%
**Student** | 92.94% | 92.94% | 0%
**React Native** | 92.71% | 92.72% | -0.01%
<p align="right">
</p>

rn/Teacher/src/modules/app/index.js
```diff
@@ -40,7 +41,7 @@ const teacher = {
 
 const student = {
   appId: 'student',
-  filterCourse: (course: Course): boolean => true,
+  filterCourse: (course: Course): boolean => !course.access_restricted_by_date,
```
armstrongnate: @shichongrui are you okay with this? It could possibly deviate from the dashboard cards api but I don't think we ever want to include these courses anywhere.

We talked about removing the `filterCourse` check from Dashboard (except we need it for Teacher). Turns out Dashboard uses `filterCourse` in a bunch of places so that got out of hand real quick.

shichongrui: Yeah I think considering the usage of `filterCourse` it will be ok. Hopefully the dashboard api never returns courses that are access restricted cause that would seem broken.

PR #191, opened 2019-05-16
Fix tests with XCUITest

armstrongnate:
[ignore-commit-lint]

* XCUITest's `isHittable` will only ever be true if the element has `isAccessibilityElement = true` and by default it is `false` for `UIImageView`s
* We have to `.tap()` text inputs before we can `typeText()`


inst-danger:
**Core** | 93.54% | 93.54% | 0%
**Student** | 92.94% | 92.94% | 0%
**React Native** | 92.72% | 92.72% | 0%
<p align="right">
</p>

PR #216, opened 2019-05-23
Optimize UI tests (for xcuitest)

armstrongnate:
[ignore-commit-lint]

We don't want to wait for things that are not visible/enabled to be
not visible/enabled because we end up waiting the entire duration.

Relies on https://github.com/instructure/SwiftUITest/pull/5

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,3 +1,3 @@
 binary "https://customers.pspdfkit.com/carthage/mq7ic2x5FCWWUCuxWct2g87YfR9hpK/pspdfkit.json" "8.3.1"
 github "google/EarlGrey" "913fa9802585fe2739aeb99084aa28cd514aaf72"
-github "instructure/SwiftUITest" "ca28f10a3ca0b7c5e700076d435c507e501251b6"
+github "instructure/SwiftUITest" "7215ae92d1d1248cfa366d2bbbd6e7100840d6e4"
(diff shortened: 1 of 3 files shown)


vanwagonet:
Did this make a significant impact on the speed when using the XCUITest driver? I'm wondering if rather than making `*Now` properties, if we just always want this behavior. It may be a little more effort in a test case to have explicit waits for things, but those are already in our tests.

The typical pattern is `setup -> wait -> assert+ (-> action -> wait -> assert+)*`.

armstrongnate:
@vanwagonet Yes, it made a significant impact on the speed of the xcuitest driver. I noticed that XCUITest does not wait the full timeout in `waitForExistence` so its use is not a performance issue. The problem was that we were waiting for something that doesn't exist to not exist.. which would cause `waitForExistence` to wait the entire timeout.

I agree that removing all the waits from SwiftUITest might be the direction that we want to go but I don't think it's pressing since `waitForExistence` is optimized to bail as soon as it discovers the element exists.

We might find that having the waits in SwiftUITest is convenient, especially with live data. TBD. We might very well soon decide to switch to the model you described. For now, this is a small step toward optimization.

armstrongnate:
I'm perfectly fine with not merging this until all of my research is done and we decide on a direction. I mostly didn't want these changes to get lost. But it isn't much and is subject to change.

inst-danger:
**Core** | 95.04% | 95.04% | 0%
**Student** | 92.34% | 92.34% | 0%
<p align="right">
</p>

PR #225, opened 2019-05-28
Fix a crash on startup

armstrongnate:
[ignore-commit-lint]

I got a `BAD_ACCESS` error here. It originated from trying to set up
push notifications in `getNotificationPreferencesSetup`.

I'm not positive that this will fix the crash but it seems like we
should not strongly retain `self` here.

I tested areas of the app that use this ThreeLegit stuff and they
appear to still be working.


inst-danger:
**Core** | 95.14% | 95.16% | -0.01%
<p align="right">
</p>

PR #240, opened 2019-05-29
Fix popovers on iPad

armstrongnate:
[ignore-commit-lint]

### DIFF
--- a/Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
+++ b/Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
@@ -80,13 +80,13 @@ class SubmissionButtonPresenter: NSObject {
         let types = assignment.submissionTypes
         let arc = types.contains(.online_upload) && arcID != .pending && arcID != .none
         if !arc && types.count == 1, let type = types.first {
-            return submitType(type, for: assignment)
+            return submitType(type, for: assignment, button: button)
         }
         let alert = SubmissionButtonAlertView.chooseTypeAlert(self, assignment: assignment, arc: arc, button: button)
         view?.present(alert, animated: true, completion: nil)
     }
 
-    func submitType(_ type: SubmissionType, for assignment: Assignment) {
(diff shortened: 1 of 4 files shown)


Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
```diff
@@ -184,8 +184,8 @@ extension SubmissionButtonPresenter: FilePickerControllerDelegate {
 
 // MARK: - media_recording
 extension SubmissionButtonPresenter: AudioRecorderDelegate, UIImagePickerControllerDelegate, UINavigationControllerDelegate {
-    func pickMediaRecordingType() {
-        let alert = SubmissionButtonAlertView.chooseMediaTypeAlert(self)
+    func pickMediaRecordingType(button: UIView) {
+        let alert = SubmissionButtonAlertView.chooseMediaTypeAlert(self, button: button)
```
bluwave: `SubmissionButtonAlertView`, and `button` all sound like things that should be in the viewController and not the presenter.  Can you push the `SubmissionButtonAlertView.chooseMediaTypeAlert` call down to the viewController?  You have the presenter there and can pass that in?  Not a big deal.  Maybe this is a knit pick.  Thoughts?

inst-danger:
**Core** | 95.16% | 95.16% | 0%
<p align="right">
</p>

Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
armstrongnate: Yeah I agree with you but we refactored this stuff out of the view controller because the submit button can live in multiple view controllers. So if we remove it from here (one place) we will have to duplicate it in two different places (assignment details and submission details view controllers).

You can see we do the very same thing here:

https://github.com/instructure/canvas-ios/blob/b8ff6eda444931489e8e667d561527e3e834cdf2/Student/Student/Submissions/SubmissionButton/SubmissionButtonAlertView.swift#L38-L40

I can move it to the vc but just want us to be aware that we will have to duplicate it in multiple VCs.

PR #250, opened 2019-05-29
Remove Persistence protocol

vanwagonet:
refs: MBL-12080
affects: none
release note: none


vanwagonet:
Student presenter tests keep failing in bitrise. :( I've tried a few different fixes, but I'm not sure I understand the root cause.

armstrongnate:
I have an idea. I'll push it up in a sec.

armstrongnate:
I think that using expectations in that way still allows for race conditions. I think using callbacks like I have in https://github.com/instructure/canvas-ios/pull/250/commits/dce417b5a18c3cae9a69a539f4eafadbfd855784 will be more stable. We'll see what the builds say.

armstrongnate:
The root problem is that the expectations get called multiple times, so we would have to know exactly how many times they need to get called and then set `expectedFulfillmentCount` but that is extremely brittle.

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: None

[MBL-12080](https://instructure.atlassian.net/browse/MBL-12080)

**Core** | 95.13% | 95.16% | -0.03%
**React Native** | 92.68% | 92.68% | 0%
<p align="right">
</p>

PR #271, opened 2019-06-05
Make RN errors in crashlytics slightly better

shichongrui:
refs: MBL-12419
affects: Student,Teacher
release note: none


rn/Teacher/src/routing/index.js
```diff
@@ -78,7 +79,14 @@ export function wrapComponentInProviders (moduleName: string, generator: (props:
...
         global.crashReporter.setBool('screenError', true)
-        global.crashReporter.recordError(error)
+
+        let domain = `com.instructure.ios.${app.current().appId}`
+        global.crashReporter.recordError({
+          domain,
+          userInfo: {
+            errorMessage: error.message.split('\n')[0]
+          }
+        })
```
shichongrui: The big changes are the domain including the app id and the `userInfo`, which will show up in the `keys` section of an error in crashlytics, including the first line of the error message so we don't have to dig into the logs trying to find it.

vanwagonet: Will this also help separate different errors? I think we have an issue with RN bugs all getting lumped together.

shichongrui: Unfortunately no. I tried to get that to work but couldn't come up with a good way. The way firebase groups errors is by the domain and code. I tried making the domain based on the first line of the error message but it seems that it's not a valid domain and thus doesn't get recorded. So then I tried using the name of the component that errors but in production everything is minified so it's useless. The last option would be to use the `moduleName` but I didn't try that because I figured the url would not be a valid domain as well. If you have any thoughts for a potential option that could make this work I'd love to make this a thing.

vanwagonet: If code can be any string, then the `moduleName` could be the code. or the error message. or a digest of the message.

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


[MBL-12419](https://instructure.atlassian.net/browse/MBL-12419)

**Core** | 95.15% | 95.14% | 0.01%
**React Native** | 92.68% | 92.67% | 0%
<p align="right">
</p>

rn/Teacher/src/routing/index.js
shichongrui: It looks like the code has to be a number.

PR #277, opened 2019-06-05
Fix masquerading as user in a different locale

armstrongnate:
[ignore-commit-lint]

Switching locale prompts a needed alert that would get dismissed. We
don't need to dismiss since we animate to a new root view controller
when we masquerade.

pork: remove auto correct from domain field in masquerade screen.


inst-danger:
**Core** | 95.15% | 95.15% | 0%
**React Native** | 92.67% | 92.67% | 0%
<p align="right">
</p>

PR #293, opened 2019-06-11
Improve grade slider a11y

shichongrui:
refs: MBL-12487
affects: Teacher
release note: Improved speed grade accessibility

This PR also includes a button on the dashboard to take you back to the last screen
you visited because I got sick of going into the developer menu finding the routes.

### DIFF
--- a/Canvas/Canvas/CanvasAppDelegate+Helm.swift
+++ b/Canvas/Canvas/CanvasAppDelegate+Helm.swift
@@ -41,7 +41,7 @@ extension AppDelegate: RCTBridgeDelegate {
                 self.changeUser()
                 return	
             }
-            self.setup(session: session, wasReload: ProcessInfo.processInfo.arguments.contains("-Reload"))
+            self.setup(session: session, wasReload: true)
         }
 
         // Files
--- a/Canvas/Canvas/CanvasAppDelegate+Helm.swift
+++ b/Canvas/Canvas/CanvasAppDelegate+Helm.swift
@@ -37,7 +37,11 @@ extension AppDelegate: RCTBridgeDelegate {
 
         HelmManager.shared.onReactReload = {
             guard self.window?.rootViewController is CanvasTabBarController else { return }
-            self.changeUser()
+            guard let session = Keychain.mostRecentSession else {
+                self.changeUser()
+                return	
+            }
+            self.setup(session: session, wasReload: ProcessInfo.processInfo.arguments.contains("-Reload"))
         }
 
         // Files
--- a/rn/Teacher/index.ios.js
+++ b/rn/Teacher/index.ios.js
@@ -43,6 +43,7 @@ import Navigator from './src/routing/Navigator'
 import { featureFlagSetup } from './src/common/feature-flags'
 import APIBridge from './src/canvas-api/APIBridge'
 import { Crashlytics } from './src/common/CanvasCrashlytics'
+import { getLastRoute } from './src/modules/developer-menu/DeveloperMenu'
 
 global.crashReporter = Crashlytics
 
@@ -72,6 +73,7 @@ const loginHandler = async ({
   skipHydrate,
   countryCode,
   locale,
+  wasReload,
 }: {
   appId: AppId,
   authToken: string,
@@ -82,6 +84,7 @@ const loginHandler = async ({
   skipHydrate: boolean,
   countryCode: string,
   locale: string,
+  wasReload: boolean,
 }) => {
   setupI18n(locale || NativeModules.SettingsManager.settings.AppleLocale)
   App.setCurrentApp(appId)
@@ -137,6 +140,12 @@ const loginHandler = async ({
   Helm.loginComplete()
   loginVerify()
   beginUpdatingBadgeCounts()
+
+  if (wasReload) {
+    let lastRoute = await getLastRoute()
+    let navigator = new Navigator('')
+    navigator.show(lastRoute.url, lastRoute.options, lastRoute.props)
+  }
 }
 
 const emitter = new NativeEventEmitter(NativeLogin)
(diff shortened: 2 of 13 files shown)


inst-danger:
#### Release Note: 
Improved speed grade accessibility

#### Affected Apps: Teacher

[MBL-12487](https://instructure.atlassian.net/browse/MBL-12487)

**Core** | 95.16% | 95.16% | 0%
**React Native** | 92.69% | 92.7% | -0.02%
<p align="right">
</p>

rn/Teacher/src/modules/speedgrader/components/Slider.js
```diff
@@ -239,6 +239,24 @@ export default class Slider extends Component<Props, State> {
...
 
+  sliderAccessibilityLabel () {
+    if (this.props.excused) {
+      return i18n('Excused')
+    } else if (this.props.score !== null) {
+      return i18n(`{ score, number } out of { pointsPossible, number } {
+        points, plural,
+             =1 {point}
+          other {points}
+      } possible`, {
```
vanwagonet: `{pointsPossible, number} {points, plural}` seems like an odd choice, when they can be easily combined. Also the keyword `one` isn't the same thing as `=1`. Since we are mapping to the linguistic singular, `one` would be better here.

```
... out of { pointsPossible, plural,
    one {# point possible}
  other {# points possible}
}
```

rn/Teacher/src/modules/dashboard/Dashboard.js
```diff
@@ -437,7 +452,8 @@ export class Dashboard extends React.Component<Props, State> {
           accessibilityLabel: i18n('Edit Dashboard'),
           action: this.showFavoritesList,
           disabled: !this.props.totalCourseCount || Boolean(this.props.pending),
-        }],
+        },
+        ...devOnlyButtons],
```
vanwagonet: Is all this just to avoid dev tapping into somewhere? Would a temporary scheme launch argument be better?

rn/Teacher/src/modules/speedgrader/components/Slider.js
shichongrui: Ohh I forgot that you could do that. Great thanks!

rn/Teacher/src/modules/dashboard/Dashboard.js
shichongrui: I was getting tired of having to tap either through the app or into the developer menu every time I wanted to check a change so I added this button for development only to take me to the last screen I was on. I can remove if others don't want this but this was a life saver for me from having to go into speed grader every time.

shichongrui: I personally dislike launch arguments because I always forget to remove them.

vanwagonet: I can be ok with this if it makes your life easier. ;)

Canvas/Canvas/CanvasAppDelegate+Helm.swift
```diff
@@ -37,7 +37,11 @@ extension AppDelegate: RCTBridgeDelegate {
 
         HelmManager.shared.onReactReload = {
             guard self.window?.rootViewController is CanvasTabBarController else { return }
-            self.changeUser()
+            guard let session = Keychain.mostRecentSession else {
+                self.changeUser()
+                return	
+            }
+            self.setup(session: session, wasReload: ProcessInfo.processInfo.arguments.contains("-Reload"))
```
shichongrui: I'm not sure what the reason was for doing `changeUser` instead of taking us right back into the same logged in user on reload so if this is incorrect let me know.

rn/Teacher/src/modules/speedgrader/components/Slider.js
```diff
@@ -258,6 +257,7 @@ export default class Slider extends Component<Props, State> {
   }
 
   render () {
+    console.log(this.sliderAccessibilityLabel())
```
vanwagonet: debug

Canvas/Canvas/CanvasAppDelegate+Helm.swift
vanwagonet: I think this is a good change. The important thing was making sure that reset the root view controller.

I'm not sure I follow what's going on with the command line argument though. Wouldn't we always pass true from here?

rn/Teacher/index.ios.js
```diff
@@ -137,6 +140,12 @@ const loginHandler = async ({
   Helm.loginComplete()
   loginVerify()
   beginUpdatingBadgeCounts()
+
+  if (wasReload) {
+    let lastRoute = await getLastRoute()
+    let navigator = new Navigator('')
+    navigator.show(lastRoute.url, lastRoute.options, lastRoute.props)
```
vanwagonet: this will bring you back to the same screen, but it won't have the navigation stack below. Is that ok?

Canvas/Canvas/CanvasAppDelegate+Helm.swift
shichongrui: I've switched to a launch argument. If the launch argument is enabled this will return true, if not it will return false.

rn/Teacher/index.ios.js
shichongrui: Yeah, for now that works for me. If we want to try to get the whole navigation stack at some point we could investigate that.

rn/Teacher/src/modules/speedgrader/components/Slider.js
shichongrui: 👍

Canvas/Canvas/CanvasAppDelegate+Helm.swift
vanwagonet: Is that so we'll only get the js back to my spot if you opt in?

shichongrui: Correct. CMD-Option-R will open launch arguments and you can toggle it.

vanwagonet: So why not just always do that? I'd rather not have more options around. Reload is already a dev feature.

PR #376, opened 2019-07-10
Fix file uploads

armstrongnate:
[ignore-commit-lint]

### DIFF
--- a/Canvas/SubmitAssignment/SubmitAssignmentPresenter.swift
+++ b/Canvas/SubmitAssignment/SubmitAssignmentPresenter.swift
@@ -122,9 +122,6 @@ class SubmitAssignmentPresenter {
                 .appendingPathComponent(url.lastPathComponent)
             do {
                 try url.move(to: newURL)
-                if let image = UIImage(contentsOfFile: newURL.path), let data = image.normalize().jpegData(compressionQuality: 0.8) {
-                    try data.write(to: newURL)
-                }
                 urls.append(newURL)
                 self.load(attachments: attachments, into: urls, callback: callback)
             } catch {
--- a/Core/Core/API/API.swift
+++ b/Core/Core/API/API.swift
@@ -38,7 +38,6 @@ public struct URLSessionAPI: API {
     public let actAsUserID: String?
     public let baseURL: URL
     let urlSession: URLSession
-    public let sharedContainerIdentifier: String?
 
     public var identifier: String? {
         return urlSession.configuration.identifier
@@ -58,14 +57,12 @@ public struct URLSessionAPI: API {
         accessToken: String? = nil,
         actAsUserID: String? = nil,
         baseURL: URL? = nil,
-        urlSession: URLSession = URLSessionAPI.defaultURLSession,
-        sharedContainerIdentifier: String? = Bundle.main.appGroupID()
+        urlSession: URLSession = URLSessionAPI.defaultURLSession
     ) {
         self.accessToken = accessToken
         self.actAsUserID = actAsUserID
         self.baseURL = baseURL ?? URL(string: "https://canvas.instructure.com/")!
         self.urlSession = urlSession
-        self.sharedContainerIdentifier = sharedContainerIdentifier
     }
 
     @discardableResult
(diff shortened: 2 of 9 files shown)


Canvas/SubmitAssignment/SubmitAssignmentPresenter.swift
```diff
@@ -122,9 +122,6 @@ class SubmitAssignmentPresenter {
                 .appendingPathComponent(url.lastPathComponent)
             do {
                 try url.move(to: newURL)
-                if let image = UIImage(contentsOfFile: newURL.path), let data = image.normalize().jpegData(compressionQuality: 0.8) {
```
armstrongnate: `UIImage.normalize()` uses too much memory for share extensions. We'll have to normalize on the backend or via PSPDFKit or something.

inst-danger:
**Core** | 95.34% | 95.34% | -0%
**Student** | 90.78% | 90.78% | 0%
**React Native** | 92.78% | 92.78% | 0%
<p align="right">
</p>

Core/Core/API/API.swift
```diff
@@ -101,7 +101,8 @@ public struct URLSessionAPI: API {
     @discardableResult
     public func uploadTask<R: APIRequestable>(_ requestable: R) throws -> URLSessionTask {
         let request = try requestable.urlRequest(relativeTo: baseURL, accessToken: accessToken, actAsUserID: actAsUserID)
-        let url = URL.temporaryDirectory.appendingPathComponent(UUID.string)
+        let directory = urlSession.configuration.sharedContainerIdentifier.flatMap(URL.sharedContainer) ?? URL.temporaryDirectory
```
vanwagonet: since we don't always write to a temp file, should we make sure to delete the local file when we're done?

armstrongnate: Done.

Core/Core/Use Cases/UploadManager.swift
```diff
@@ -215,32 +198,81 @@ open class UploadManager: NSObject, URLSessionDelegate, URLSessionTaskDelegate,
...
+                guard let data = file.uploadData else {
+                    complete(file: file, error: error)
+                    break
+                }
+                do {
+                    let result = try decoder.decode(APIFile.self, from: data)
+                    file.update(fromAPIModel: result)
+                    try context.save()
+                    if case let .submission(courseID, assignmentID, comment)? = file.context, error == nil {
+                        submit(file: file, courseID: courseID, assignmentID: assignmentID, comment: comment)
```
bluwave: do you need to call `complete` in this case?

armstrongnate: No, `complete` will cause the file to disappear from the uploading screen. I want it to hang around until the submission request finishes just in case that request fails (or takes a long time).

armstrongnate: Good call out though.

Core/Core/Use Cases/UploadManager.swift
```diff
@@ -215,32 +198,81 @@ open class UploadManager: NSObject, URLSessionDelegate, URLSessionTaskDelegate,
...
                 let file = self.file(taskID: task.taskIdentifier)
             else { return }
+            let decoder = JSONDecoder()
+            decoder.dateDecodingStrategy = .iso8601
             switch step {
             case .upload:
-                if case let .submission(courseID, assignmentID, comment)? = file.context, error == nil {
-                    submit(file: file, courseID: courseID, assignmentID: assignmentID, comment: comment)
-                    return
+                // On device the upload might fail with a 201
```
bluwave: I'm not too familiar with File Uploads so take this with a grain of salt.  This method seems very large and to have a ton of different logic in it.  It appears you tested it well. Would it be helpful or possible (maybe it's not helpful or possible :shrug:) to break this into smaller chunk size methods and easier to map the testing to individual methods?  Just a thought.

Core/Core/Use Cases/UploadManager.swift
bluwave: cool cool cool

Core/Core/Use Cases/UploadManager.swift
armstrongnate: I agree it's pretty difficult to follow. I will break it up.

armstrongnate: I'm also going to see if the `fail with a 201` is still a thing. It only happened on Bitrise build so I'll have to wait for it to build.

armstrongnate: I pushed up a refactor. I think I still prefer always testing through `urlSession(_:task:didCompleteWithError:)` since that is the entry point but I already like having that function split up and chunked.

Thanks for the feedback.

armstrongnate:
I am seeing the 201 error thing with adhoc build. I'll reapply the fix.

PR #401, opened 2019-07-19
MBL-12849 - Refactor PageViewAnalytics - Move to Core

bluwave:
refs: MBL-12849
affects: Student
release note: none

### DIFF
--- a/Core/Core/Keychain/GeneralPurposeKeychain.swift
+++ b/Core/Core/Keychain/GeneralPurposeKeychain.swift
@@ -39,8 +39,12 @@ class GeneralPurposeKeychain {
 
     @discardableResult
     func setItem(_ value: Any, for key: String) -> Bool {
-        let data = NSKeyedArchiver.archivedData(withRootObject: value)
-        return setData(data, for: key)
+        do {
+            let data = try NSKeyedArchiver.archivedData(withRootObject: value, requiringSecureCoding: false)
+            return setData(data, for: key)
+        } catch {
+            return false
+        }
     }
 
     @discardableResult
--- a/Core/Core/Keychain/GeneralPurposeKeychain.swift
+++ b/Core/Core/Keychain/GeneralPurposeKeychain.swift
@@ -38,16 +38,6 @@ class GeneralPurposeKeychain {
     }
 
     @discardableResult
-    func setItem(_ value: Any, for key: String) -> Bool {
-        do {
-            let data = try NSKeyedArchiver.archivedData(withRootObject: value, requiringSecureCoding: false)
-            return setData(data, for: key)
-        } catch {
-            return false
-        }
-    }
-
-    @discardableResult
     func setData(_ data: Data, for key: String) -> Bool {
         var query: [CFString: Any] = [
             kSecClass: kSecClassGenericPassword,
@@ -96,14 +86,6 @@ class GeneralPurposeKeychain {
             return result as? Data
         }
         return nil
-    }
-
-    func item(for key: String) -> Any? {
-        guard let data: Data = data(for: key) else {
-            return nil
-        }
-
-        return NSKeyedUnarchiver.unarchiveObject(with: data) as? NSCoding
     }
 
     @discardableResult
--- a/Core/Core/Keychain/GeneralPurposeKeychain.swift
+++ b/Core/Core/Keychain/GeneralPurposeKeychain.swift
@@ -1,0 +1,119 @@
+//
+// This file is part of Canvas.
+// Copyright (C) 2019-present  Instructure, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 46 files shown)


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Student

[MBL-12849](https://instructure.atlassian.net/browse/MBL-12849)

**Core** | 95.28% | 95.38% | -0.1%
**Student** | 90.78% | 90.78% | -0.01%
**React Native** | 92.78% | 92.78% | 0%
<p align="right">
</p>

Core/Core/Keychain/GeneralPurposeKeychain.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    init(serviceName: String, accessGroup: String? = nil) {
+        self.serviceName = serviceName
+        self.accessGroup = accessGroup
+    }
+
+    @discardableResult
+    func setItem(_ value: Any, for key: String) -> Bool {
+        let data = NSKeyedArchiver.archivedData(withRootObject: value)
```
vanwagonet: Do we need this to be NSCoding based? Or could we make use `Encodable` instead of of `Any`?

Core/Core/PageViewAnalytics/PageViewEventViewControllerLoggingProtocol.swift
```diff
@@ -18,16 +18,20 @@
 
 import Foundation
 
+// swiftlint:disable:next private_over_fileprivate
 fileprivate var tagAssociationStartKey: UInt8 = 0
+// swiftlint:disable:next private_over_fileprivate
```
vanwagonet: why not use `private` instead of disabling the lint?

Student/Student/Assignments/AssignmentDetails/AssignmentDetailsPresenter.swift
```diff
@@ -36,7 +36,13 @@ protocol AssignmentDetailsViewProtocol: SubmissionButtonViewProtocol {
     func showSubmitAssignmentButton(title: String?)
 }
 
-class AssignmentDetailsPresenter {
+class AssignmentDetailsPresenter: PageViewLoggerPresenterProtocol {
+
+    var pageViewEventName: String {
+        guard let a = assignments.first else { return "\(String(describing: self))"}
```
vanwagonet: seems like it would happen often that the viewDidAppear will happen before the assignment is loaded. We have thr courseID & assignmentID as well as the baseURL from the session. Perhaps it would be better if we always constructed the url, instead of relying on the core data object.

TestsFoundation/TestsFoundation/API/MockAPI.swift
```diff
@@ -56,7 +56,10 @@ public class MockAPI: API {
         if let (data, response, error) = mocks[request] {
             var value: R.Response?
             if let data = data {
-                value = try! JSONDecoder().decode(R.Response.self, from: data)
+                value = try? requestable.decode(data)
```
vanwagonet: This is probably what we wanted all along. Do we have cases that rely on the generic JSONDecoder below, or could we just remove that?

Student/Student/Assignments/AssignmentDetails/AssignmentDetailsPresenter.swift
```diff
@@ -87,14 +93,22 @@ class AssignmentDetailsPresenter {
...
     }
+    var pageViewLogger: PageViewEventViewControllerLoggingProtocol
 
-    init(env: AppEnvironment = .shared, view: AssignmentDetailsViewProtocol, courseID: String, assignmentID: String, fragment: String? = nil) {
+    init(env: AppEnvironment = .shared,
+         view: AssignmentDetailsViewProtocol,
+         courseID: String,
+         assignmentID: String,
+         fragment: String? = nil,
+         pageViewLogger: PageViewEventViewControllerLoggingProtocol = PresenterPageViewLogger()
```
vanwagonet: I don't love having one more thing to pass into each of these presenters. Could the logger be on the AppEnvironment?

Student/Student/Assignments/AssignmentDetails/AssignmentDetailsPresenter.swift
bluwave: great suggestion

Student/Student/Assignments/AssignmentDetails/AssignmentDetailsPresenter.swift
bluwave: this is a good suggestion.  I like this better as well.  Fixed.

Core/Core/Keychain/GeneralPurposeKeychain.swift
bluwave: My thinking was that NSCoding seems to handle dictionaries easier.  Also, with `Encodble` i get that ```Protocol type 'Encodable' cannot conform to 'Encodable' because only concrete types can conform to protocols``` dance.

vanwagonet: Yeah, you have to use a generic constraint for Encodable. `setItem<V: Encodable>(_ value: V,`

My concern is can't archivedData fail, or can it really archive `Any`? It doesn't look like it throws, so does that mean it just crashes?

bluwave: I've added a try catch

PR #436, opened 2019-07-26
Remove old SubmitAssignment share action

armstrongnate:
Enhances new SubmitAssignment extension to select default course
and assignment.

refs: MBL-12871
affects: student
release note: none


Canvas/TechDebt/Files and Folders/FileViewController.m
```diff
@@ -153,7 +153,7 @@ - (void)applyRoutingParameters:(NSDictionary *)params {
     if (courseID) {
         self.courseID = [NSNumber numberWithInteger:courseID.integerValue];
     }
-    NSString *assignmentID = params[@"query"][@"courseID"];
+    NSString *assignmentID = params[@"query"][@"assignmentID"];
```
armstrongnate: 😆

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Student

[MBL-12871](https://instructure.atlassian.net/browse/MBL-12871)

**Core** | 95.33% | 95.34% | -0%
**Student** | 91.14% | 91.14% | 0%
**React Native** | 92.79% | 92.79% | 0%
<p align="right">
</p>

shichongrui:
Code looks good but looks like one of the SubmitAssignmentPresenterTests failed.

Canvas/SubmitAssignment/SubmitAssignmentPresenter.swift
```diff
@@ -67,10 +71,26 @@ class SubmitAssignmentPresenter {
         env = AppEnvironment.shared
         env.userDidLogin(session: session)
         sharedContainer = container
+
+        if let courseID = env.userDefaults?.submitAssignmentCourseID {
```
bluwave: when does this get set?

armstrongnate: https://github.com/instructure/canvas-ios/pull/436/files#diff-a9a597152bf01b148f87e0546fc56aceR41

armstrongnate: This is when they are annotating a file in our app and then want to submit their annotations using our submit share extension.

Core/Core/AppEnvironment/SessionDefaults.swift
```diff
@@ -0,0 +1,59 @@
...
+    private var userDefaults: UserDefaults {
+        return UserDefaults(suiteName: Bundle.main.appGroupID()) ?? .standard
+    }
+
+    private var sessionDefaults: [String: Any]? {
+        get { return userDefaults.dictionary(forKey: session.uniqueID) }
+        set { userDefaults.set(newValue, forKey: session.uniqueID) }
+    }
+
+    public let session: KeychainEntry
```
bluwave: would it be better to use `AppEnvironment.currentSession` here instead of adding another session variable?

bluwave: oh wait, i think im seeing how this works, what about passing in the session.uniqueID?

Core/Core/Keychain/Keychain.swift
```diff
@@ -54,6 +54,17 @@ public struct KeychainEntry: Codable, Hashable {
         return masquerader?.host.flatMap { URL(string: "https://\($0)") }
     }
 
+    public var uniqueID: String {
```
bluwave: This feels a lot like a hashValue.  Have you considered just using `hashValue` of the KeychainEntry?

armstrongnate: I thought about using `hashValue` but `Hashable` isn't guaranteed to be the same across executions because they use a random seed each time. Docs say this:

```
    /// Hash values are not guaranteed to be equal across different executions of
    /// your program. Do not save hash values to use during a future execution.
```

That makes me think it's not a great fit for UserDefaults since they are persisted across executions.

Core/Core/AppEnvironment/SessionDefaults.swift
armstrongnate: Good call. I don't need the entire entry- I just need the uniqueID. I will switch to only taking that instead.

PR #439, opened 2019-07-26
MBL-12901 - make core safe for extensions

bluwave:
refs: MBL-12901
affects: Student
release note: none


armstrongnate:
Whoa dang that's a lot of build failures. Was Bitrise busted or something?

bluwave:
no, these are legit, working through them

PR #463, opened 2019-08-01
do not use packager for debug builds

bluwave:
this should be fairly easy to put back to only use packager for debug builds, just take out the checks for `Debug` if you find yourself needing it.

refs: none
affects: Student
release note: none

### DIFF
--- a/Canvas/Canvas.xcodeproj/project.pbxproj
+++ b/Canvas/Canvas.xcodeproj/project.pbxproj
@@ -4095,7 +4095,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Student

**Core** | 94.97% | 94.97% | 0%
**Student** | 92.97% | 92.97% | 0%
**React Native** | 92.76% | 92.76% | 0%
<p align="right">
</p>

vanwagonet:
Why not use the packager? Doesn't this make the problem of rebuilds much worse for dev?

bluwave:
I find myself turning this off quite often lately to install on device.  Seems we do far less RN work these days as well.  If we are against this,  No worries at all, we can keep it in, I can be neutral on this.

bluwave:
👍

armstrongnate:
An environment variable set in the scheme settings might be a little more user friendly (so that it's a checkbox) but ultimately I'm okay with this.

bluwave:
@armstrongnate - i looked into an environment variable and could not get access to the variable in a script phase.  Have you seen that work before?  Perhaps i'm missing something.

armstrongnate:
I haven't seen it work.. I assumed the script would have access to those env vars. Oh well 🤷‍♂

PR #496, opened 2019-08-09
Handle enhanced RCE changes

armstrongnate:
refs: MBL-12843
affects: student, teacher
release note: none


Core/Core/Extensions/NSPersistentContainerExtensions.swift
```diff
@@ -31,15 +31,26 @@ extension NSPersistentContainer {
         }
 
         container.loadPersistentStores { _, error in
-            if let error = error {
-                fatalError(error.localizedDescription)
+            guard error == nil else {
+                container.destroy() // ignore migration conflicts
```
bluwave: Do you think it would be helpful to try to delete any existing`.sql` files as well?  Not sure if that would be helpful or not.  I read somewhere about errors in `container.loadPersistenstores` being because of a corrupt sql file.  I'll have to dig up the link.

bluwave:
this looks good to me.  Lots of js in there.  May be helpful to have @vanwagonet look at it as well.

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


[MBL-12843](https://instructure.atlassian.net/browse/MBL-12843)

**Canvas iOS** | 94.83% | 94.88% | -0.06%
<p align="right">
</p>

Core/Core/Extensions/NSPersistentContainerExtensions.swift
armstrongnate: Yeah I think I know what you're talking about. Something to do with the journal files right? I think that would be worth looking into at some point.

armstrongnate:
@vanwagonet How we've handled this in the past (and our current plan for this feature) is to enable the feature with a code change and a release after it goes to prod on Canvas. We do end up with a time period where the feature is broken but that's why we need smarter feature flags. 😉

armstrongnate:
@vanwagonet your thoughts have created some discussion. I'm going to hold off on this for now until we're sure about the plan for this feature flag.

In the meantime, I'm going to pull out the database changes in this pr and apply them separately so that I can cut a release with just that change.

Core/Core/Use Cases/GetEnabledFeatureFlags.swift
```diff
@@ -0,0 +1,58 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Foundation
+import CoreData
+
+public class GetEnabledFeatureFlags: APIUseCase {
```
vanwagonet: I think this needs to be a CollectionUseCase to get the deleting of prior list before saving the new list.

armstrongnate:
This isn't ready yet. I need to update our media upload stuff to post to canvas after the upload finishes.

Core/Core/Use Cases/GetEnabledFeatureFlags.swift
armstrongnate: Good call! Thanks.

armstrongnate:
Nice catch, @vanwagonet. Should be ready for re-review now.

PR #510, opened 2019-08-19
Don't return deleted objects from store

armstrongnate:
refs: MBL-13082
affects: student, teacher
release note: Fixed a crash when viewing pages list


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1003859&projectid=17)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

instructure-cx:
Scan submitted to Checkmarx

armstrongnate:
I think this is what we want. I think there's a short window where deleted objects are in the FRC. By returning nil from the subscript, we're not messing with any of the section or row counts. If a consumer of the store encounters a nil object for an indexPath they can assume the object has been deleted (and should be getting a call to update very soon).

Core/Core/Store/Store.swift
```diff
@@ -104,11 +104,15 @@ public class Store<U: UseCase>: NSObject, NSFetchedResultsControllerDelegate {
...
+        }
+        return object
     }
 
     public subscript(index: Int) -> U.Model? {
-        return frc.object(at: IndexPath(row: index, section: 0))
+        return self[IndexPath(row: index, section: 0)]
     }
 
     public func numberOfObjects(inSection section: Int) -> Int {
```
bluwave: does this method still work?

bluwave: would this still cause a tableview to have say x number of rows but if one is nil, then that count would be off by 1?

armstrongnate: Yeah, the counts are all still the same. The FRC is including deleted objects in these counts during certain race conditions. So the object is there, it's just that all its properties are faults and `.isDeleted` is true.

The problem is that when we try to access non-optional properties that are faults we get a crash.

I guess this is one reason we _should_ make all NSManaged properties optional but yuck.

armstrongnate: > would this still cause a tableview to have say x number of rows but if one is nil, then that count would be off by 1?

No, it shouldn't. The count methods are including these deleted objects.

There is an object at the indexPath and the count includes it, but the object has been deleted. So now when we go asking for the object we will get `nil` back so that we don't try to access properties on a deleted object.

This is where the most prevalent crash is at right now:
https://github.com/instructure/canvas-ios/blob/60296d7437690aa21bb4151e21ddd66d426384e4/Core/Core/Pages/PageList/PageListViewController.swift#L167-L174

On line 172 we pass a deleted page to `cell.update`. In that update method we access non-optional properties on the deleted page. That causes a crash because the properties are nil because the page has been deleted.

armstrongnate: I can update the test to verify the `numberOfSections` is still correct.

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1004096&projectid=17)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

armstrongnate:
@bluwave I updated the test to include checks on `numberOfSections` and `numberOfObjects(inSection:)`. If you're still hesitant about this let me know.

inst-danger:
#### Release Note: 
Fixed a crash when viewing pages list

[MBL-13082](https://instructure.atlassian.net/browse/MBL-13082)

**Canvas iOS** | 94.97% | 94.98% | -0%
<p align="right">
</p>

bluwave:
Looks good

PR #522, opened 2019-08-21
Exclude unpublished courses

armstrongnate:
[ignore-commit-lint]

### DIFF
--- a/Core/Core/API/APICourseRequestable.swift
+++ b/Core/Core/API/APICourseRequestable.swift
@@ -24,13 +24,20 @@ public struct GetCoursesRequest: APIRequestable {
 
     public enum State: String {
         case available, completed, unpublished, deleted
+
+        public static var defaults: [State] {
+            if Bundle.main.isTeacherApp {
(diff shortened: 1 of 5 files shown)


Core/Core/API/APICourseRequestable.swift
```diff
@@ -24,13 +24,20 @@ public struct GetCoursesRequest: APIRequestable {
 
     public enum State: String {
         case available, completed, unpublished, deleted
+
+        public static var defaults: [State] {
+            if Bundle.main.isTeacherApp {
```
armstrongnate: Anyone have thoughts on how to override this for tests?

vanwagonet: Maybe `isTeacherApp` should also be true if the the main bundle is the runner?

armstrongnate: Good idea. Done.

inst-danger:
**Canvas iOS** | 94.98% | 94.86% | 0.12%
<p align="right">
</p>

PR #546, opened 2019-08-27
Added UI tests for discussion replies

cobbal:
This revealed several bugs (I think) one of which is preventing me from finishing the "with attachment" test.

The broken tests are disabled here (xtest...)

refs: MBL-12751
affects: none
release note: none

### DIFF
--- a/Core/Core/API/APIDiscussionRequestable.swift
+++ b/Core/Core/API/APIDiscussionRequestable.swift
@@ -50,11 +50,11 @@ struct PostDiscussionEntryRequest: APIRequestable {
     let method = APIMethod.post
     let replyId: String?
 
-    init(context: Context, topicID: ID, body: Body?, entryID: ID? = nil) {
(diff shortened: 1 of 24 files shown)


Core/Core/API/APIDiscussionRequestable.swift
```diff
@@ -37,6 +37,7 @@ struct PostDiscussionTopicRequest: APIRequestable {
 }
 
 // https://canvas.instructure.com/doc/api/discussion_topics.html#method.discussion_topics_api.add_entry
+// https://canvas.instructure.com/doc/api/discussion_topics.html#method.discussion_topics_api.add_reply
```
armstrongnate: Is this line a duplicate of the line above it?

cobbal: Nope. Different anchor, slightly different text. Felt like a waste making another datatype

PR #608, opened 2019-09-12
MBL-11243 grades list presenter and AssignmentGroups

bluwave:
MBL-11243 grades list
refs: MBL-11243
affects: Parent
release note: none


Core/Core/Models/AssigmentGroup.swift
```diff
@@ -37,15 +36,9 @@ public final class AssignmentGroup: NSManagedObject {
...
-            for a in apiAssignments {
-                let assignmentPred = NSPredicate(format: "%K == %@", #keyPath(Assignment.id), a.id.value)
-                let dbModel: Assignment = context.fetch(assignmentPred).first ?? context.insert()
-                dbModel.update(fromApiModel: a, in: context, updateSubmission: false)
-                model.assignments.insert(dbModel)
-            }
-        }
+        let assignmentPredicate = NSPredicate(format: "%K == %@", #keyPath(Assignment.assignmentGroupID), item.id.value)
+        let assignments: [Assignment] = context.fetch(assignmentPredicate)
+        assignments.forEach { $0.assignmentGroupPosition = item.position }
```
bluwave: this is the only way to get the sections in the right order

Core/Core/Grades/GradesViewController.swift
```diff
@@ -44,39 +43,39 @@ public class GradesViewController: UIViewController {
...
     }
 
     public func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return assignmentsByGroup[section].count
+        return presenter.assignments.numberOfObjects(inSection: section)
     }
 
     public func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let a = assignmentsByGroup[indexPath.section][indexPath.row]
+        let a = presenter.assignments[indexPath]
```
armstrongnate: Yes! I like this.

Core/Core/Models/AssigmentGroup.swift
armstrongnate: I think this is fine.

Core/Core/Grades/GradesPresenter.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    init(env: AppEnvironment = .shared, view: GradesViewProtocol, courseID: String, sort: GetAssignments.Sort = GetAssignments.Sort.dueAt) {
+        self.courseID = courseID
+        self.env = env
+        self.view = view
+        self.sort = sort
+    }
+
+    func viewIsReady() {
+        assignments.refresh()
```
armstrongnate: This does not exhaust and I'm not seeing `assignments.getNextPage()` when the user scrolls to the bottom. Don't we need that? I think we can also make sure we're not requesting 99 assignments but rather something more reasonable like 10 or 20 per page.

Core/Core/Grades/GradesPresenter.swift
```diff
@@ -0,0 +1,73 @@
...
+    func updateAssignmentGroups() {
+        if !assignments.pending && !didFetchGroups {
+            didFetchGroups = true
+            assignmentGroups.refresh()
+        }
+    }
+
+    func update() {
+        if didFetchGroups && !assignmentGroups.pending && !assignments.pending && assignments.count > 0 && assignmentGroups.count > 0 {
+            view?.update()
```
armstrongnate: An issue here is that there's 1 assignment and then a pull to refresh reveals that there are now 0 assignments, the table view won't get updated to remove the deleted assignment. It might be better to always call `view?.update()`?

Core/Core/Models/AssigmentGroup.swift
armstrongnate: In `Assignment.save` you could look up its AssignmentGroup and update the `assignmentGroupPosition`. That way they don't have to be requested and saved in a particular order.

Core/Core/Models/Assignment.swift
```diff
@@ -209,6 +212,40 @@ extension Assignment {
...
+            if let dueAt = dueAt, dueAt < Clock.now {
+                return NSLocalizedString("missing", comment: "")
+            } else {
+                return nil
+            }
+        }
+
+        if submission?.late ?? false {
+            return NSLocalizedString("late", comment: "")
+        } else if let s = submission, s.type != .on_paper, s.type != .none {
```
armstrongnate: I think it would be better to check `submission.submittedAt != nil` here. I think there's a `not_graded` type that you would also need to check but `submittedAt` not being nil seems like a safer bet. `SubmissionStatusLabel.js` might be a good resource.

Core/Core/Grades/GradesViewController.swift
```diff
@@ -20,40 +20,108 @@ import UIKit
...
+
+    @objc func refresh(_ control: UIRefreshControl) {
+        presenter.assignments.refresh(force: true)
+    }
+}
+
+extension GradesViewController: GradesViewProtocol {
+    func update(isLoading: Bool) {
+        tableView.reloadData()
+        loadingView.isHidden = true
```
armstrongnate: Did you mean to use `isLoading` here?

Core/Core/Models/Assignment.swift
```diff
@@ -71,7 +73,7 @@ extension Assignment {
         details = item.description
         pointsPossible = item.points_possible
         dueAt = item.due_at
-        dueAtOrder = item.due_at == nil ? "z" : "a"
+        dueAtSortNilsAtBottom = item.due_at == nil ? Date.distantFuture : item.due_at!
```
armstrongnate: `item.due_at ?? .distantFuture` to avoid a force unwrap (I know it's perfectly safe here though)

Core/Core/Models/Assignment.swift
```diff
@@ -209,6 +212,74 @@ extension Assignment {
             return .unlocked
         }
     }
+
+    public var submissionStatus: String? {
```
armstrongnate: Looks like you're always returning a non-optional `String`. Is there a difference between returning `""` vs `nil` here?

Core/Core/Use Cases/GetAssignments.swift
```diff
@@ -45,14 +51,14 @@ public class GetAssignments: CollectionUseCase {
...
-        return GetAssignmentsRequest(courseID: courseID, orderBy: orderBy)
+        return GetAssignmentsRequest(courseID: courseID, orderBy: orderBy, include: include, querySize: requestQuerySize)
     }
 
     public var scope: Scope {
         switch sort {
         case .dueAt:
             //  this puts nil dueAt at the bottom of the list
-            let a = NSSortDescriptor(key: #keyPath(Assignment.dueAtOrder), ascending: true)
+            let a = NSSortDescriptor(key: #keyPath(Assignment.dueAtSortNilsAtBottom), ascending: true)
```
armstrongnate: I'm guessing we don't need the sort descriptor below this one anymore, right?

Core/Core/Use Cases/GetAssignments.swift
```diff
@@ -90,9 +96,39 @@ public class GetSubmittableAssignments: GetAssignments {
...
+        return Scope(predicate: predicate, order: [ a, b ])
+    }
+}
+
+public class GetAssignmentsForGrades: GetAssignments {
+
+    let groupBy: GroupBy
+
+    public enum GroupBy: String {
+        case assingnmentGroup, dueAt
```
armstrongnate: Typo: `assignmentGroup`

Core/Core/Models/Assignment.swift
bluwave: ha, not sure what i was thinking, good catch

Core/Core/Grades/GradesViewController.swift
bluwave: yes

bluwave:
UI Still needs some work.

PR #612, opened 2019-09-12
Sort assignments in ext by due date

vanwagonet:
refs: MBL-13156
affects: Student
release note: Fixed order of assignments in Submit Assignment to be by due date


Core/Core/Use Cases/GetAssignments.swift
```diff
@@ -79,7 +79,7 @@ public class GetAssignments: CollectionUseCase {
 
 public class GetSubmittableAssignments: GetAssignments {
     public init(courseID: String) {
-        super.init(courseID: courseID, sort: .name)
+        super.init(courseID: courseID, sort: .dueAt)
```
vanwagonet: this gets propagated to the api call to order by due date, which should put nil last, giving us the same order we are doing in the Scope.

PR #624, opened 2019-09-16
Allow updating avatar by tapping it in drawer

vanwagonet:
refs: MBL-13191
affects: Parent
release note: Added uploading a new avatar picture from the profile drawer

### DIFF
--- a/TestsFoundation/TestsFoundation/API/MockURLSession.swift
+++ b/TestsFoundation/TestsFoundation/API/MockURLSession.swift
@@ -19,8 +19,8 @@
 import Core
 
 public class MockURLSession: URLSession {
-    public static var dataMocks: [URLRequest: MockDataTask] = [:]
-    public static var downloadMocks: [URLRequest: MockDownloadTask] = [:]
+    public static var dataMocks: [String: MockDataTask] = [:]
+    public static var downloadMocks: [String: MockDownloadTask] = [:]
 
     public typealias MockData = MockResponse<Data>
     public typealias MockDownload = MockResponse<URL>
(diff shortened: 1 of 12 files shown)


inst-danger:
#### Release Note: 
Added uploading a new avatar picture from the profile drawer

#### Affected Apps: Parent

[MBL-13191](https://instructure.atlassian.net/browse/MBL-13191)

**Canvas iOS** | 93.11% | 93.09% | 0.02%
<p align="right">
</p>

TestsFoundation/TestsFoundation/API/MockURLSession.swift
```diff
@@ -19,8 +19,8 @@
 import Core
 
 public class MockURLSession: URLSession {
-    public static var dataMocks: [URLRequest: MockDataTask] = [:]
-    public static var downloadMocks: [URLRequest: MockDownloadTask] = [:]
+    public static var dataMocks: [String: MockDataTask] = [:]
+    public static var downloadMocks: [String: MockDownloadTask] = [:]
```
bluwave: I'm probably missing it, but i'm not seeing where you are using a string in your mocking.  Why is a string vs a request needed?  What can't be turned into a request?

vanwagonet: The problem is getting two requests to be equal, or even two urls to be equal. In the case of two URLs that resolve to the same absoluteString, they still might not be equal if their baseURLs differ.

I was trying to mock the file upload and having a hard time getting them to match. The MockDistantURLSession had the same problem a while back and switched to keying off of absoluteString.

armstrongnate: I've found that getting the headers just right can make keying on URLRequest difficult. I like using absolute path because that's almost always all we care about.

PR #691, opened 2019-10-02
Fix split view bugs

armstrongnate:
refs: MBl-13352
affects: student
release note: Fixed layout issues with split view on iPad

* Trying to avoid `targetDisplayModeForAction`. It doesn't seem like we need to implement it and it's called 8 times on rotation on iOS 13. Who wants all that noise.
* Consolidating to one UISplitViewController subclass so that we don't have to maintain two
* Fixed some cases where displayModeButtonItem was not showing up when it should
* Call super in a few `viewWillTransition(to:coordinator:)` overrides because docs say to

**Test plan:**
* Tap around with iPad, expanding and collapsing split view
* Expand/collapse button should persist across navigation
* Enable and disable side-by-side and tap around the student app on iPad

PR #694, opened 2019-10-02
improve horizontal menu

bluwave:
refs: MBL-13196
affects: Parent
release note: none

test plan:

student:
1. navigate to course syllabus that has
  a. no syllabus (should only display assignments, no tabs )
  b. with syllabus (there should be two tabs)

Teacher
1. navigate to speedgrader list of students, click eye
2. make sure there are 2 tabs and you can navigate between tabs w/ iphone / ipad


instructure-cx:
Scan submitted to Checkmarx

instructure-cx:
### Checkmarx scan completed
[Full Scan Details](https://cxmgr.inssec.info/CxWebClient/ViewerMain.aspx?scanid=1007897&projectid=17)
#### Checkmarx Scan Summary
Severity|Count
---|---
High|0
Medium|0
Low|0
Informational|0

#### Violation Summary
Severity|Count
---|---

Core/Core/Views/HorizontalMenuViewController.swift
```diff
@@ -0,0 +1,278 @@
...
+            if indexPath == selectedIndexPath {
+                cell.isSelected = true
+                cell.accessibilityTraits.insert(.selected)
+            }
+
+            return cell
+        } else {
+            let identifier = "child_\(indexPath.item)"
+            let cell = collectionView.dequeueReusableCell(withReuseIdentifier: identifier, for: indexPath)
+            guard let vc = delegate?.viewControllers[indexPath.item] else { fatalError("\(#function) delegate nil") }
```
armstrongnate: Can this be an `if let` instead and don't `fatalError`? I like the `assert` above because it won't go into production but this could.

armstrongnate:
This happened when I hit the split view expand button on iPad when looking at a syllabus in Student app. It looks cool, but I don't think it's what we want.

IMG_0030

armstrongnate:
There's a black background when I pull the cell all the way. Maybe the containing view controller's view should be background lightest?

CanvasCore/CanvasCore/Helm/HelmSplitViewController.swift
```diff
@@ -97,6 +97,9 @@ extension HelmSplitViewController: UISplitViewControllerDelegate {
             top?.navigationItem.leftItemsSupplementBackButton = true
             if top?.isKind(of: EmptyViewController.self) == false {
                 top?.navigationItem.leftBarButtonItem = prettyDisplayModeButtonItem(displayMode)
+                if let observer = top as? SplitViewControllerObserverProtocol {
```
armstrongnate: Where this will fall down is with embedded view controllers.. eg if the `HorizontalMenuViewController` gets embedded in some other view controller (like a `ModuleItemDetailViewController` might seem plausible). Instead of a protocol, maybe send a notification via `NotificationCenter`?

bluwave:
@armstrongnate - i gave UIPageViewcontroller a quick try and it was not obvious to me any benefits over the current implementation.  In the quick 30 mins i gave it, was having some trouble getting at the contentOffset of the scrollview to make the slider work.  Here's one hack i did to get at the scroll view 
```
scrollView = pages.view.subviews.filter { $0 is UIScrollView }.first as? UIScrollView
```
even w/ this, the contentoffset was off a bit.

If you feel strongly, I can probably get it working, but it was not obvious to me any wins from using that different class over a collection view.  Overall, it felt more awkward.

bluwave:
> FYI: [as of iOS 9](https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver) you don't need to call `removeObserver` in deinit.
> 
> > If your app targets iOS 9.0 and later or macOS 10.11 and later, you don't need to unregister an observer in its dealloc method.

Nice, very cool.

armstrongnate:
@bluwave thanks for giving UIPageViewController a look. I'm not totally against using UICollectionView, was mostly curious if UIPageViewController would be better and sounds like it's not any better and it might even be worse. I'm fine with sticking to the current plan.

armstrongnate:
I got this to be wonky in split view. Two suggestions:

1. Wait for https://github.com/instructure/canvas-ios/pull/691 to get merged because it fixes some stuff (but I'm not sure it will fix the issues I recorded, see video below)
2. Send that notification in `splitViewController(_ svc: UISplitViewController, willChangeTo displayMode: UISplitViewController.DisplayMode)`

Here's the video:

[RPReplay_Final1570203727.MP4.zip](https://github.com/instructure/canvas-ios/files/3691270/RPReplay_Final1570203727.MP4.zip)

armstrongnate:
There's still an issue with rotation but this is so minor and quite the edge case that I don't think it's worth holding up for.

[RPReplay_Final1570218796.MP4.zip](https://github.com/instructure/canvas-ios/files/3692100/RPReplay_Final1570218796.MP4.zip)

PR #702, opened 2019-10-03
Fix page view events background task management

armstrongnate:
refs: MBL-13350
affects: student
release note: Fixed a bug that would cause the app to restart when in the background

Every time we send page view events from the background we create two
background tasks: 1 to get a token and 1 to send the events.

The problem was that we weren't calling `endBackgroundTask` for the
token task because the task was getting replaced with the second task
before we had a chance. And the system terminates apps that don't call
`endBackgroundTask`.

So instead of holding on to one task at a time we can hold on to as many
as we need.

**Test plan:**
- Launch the Student app and tap into a course
- Background the app for 30 seconds
  - No need to launch other apps. Just let your phone sit for at least 30 seconds
- Bring the Student app to the foreground by launching it
- You should be where you left off (ie the app didn't restart)

### DIFF
--- a/Core/Core/PageViewAnalytics/AppBackgroundHelperProtocol.swift
+++ b/Core/Core/PageViewAnalytics/AppBackgroundHelperProtocol.swift
@@ -20,5 +20,5 @@ import Foundation
 
 public protocol AppBackgroundHelperProtocol {
     func startBackgroundTask(taskName: String)
-    func endBackgroundTask()
+    func endBackgroundTask(taskName: String)
 }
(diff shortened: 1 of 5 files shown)


inst-danger:
#### Release Note: 
Fixed a bug that would cause the app to restart when in the background

#### Affected Apps: Student

[MBL-13350](https://instructure.atlassian.net/browse/MBL-13350)

**Canvas iOS** | 93.46% | 93.36% | 0.1%
<p align="right">
</p>

PR #706, opened 2019-10-04
Made Carthage locking mechanism more robust

cobbal:
Lock will now timeout while building instead of waiting forever

Lock will no longer be able to outlive the build process, which would
result in a permanent deadlock requiring a clean build.

[ignore-commit-lint]


inst-danger:
**Canvas iOS** | 93.45% | 94.35% | -0.9%
<p align="right">
</p>

PR #725, opened 2019-10-15
Changed drainMainQueue to be dumber, but less random

cobbal:
There just doesn't seem to be a reliable way to tell if a runloop/dispatch queue is drained. So instead just do the dumb thing.

[ignore-commit-lint]


inst-danger:
**Canvas iOS** | 93.27% | 93.27% | 0%
<p align="right">
</p>

PR #752, opened 2019-10-18
Fix marking To Do items as done

armstrongnate:
refs: MBL-13362
affects: student
release note: Fixed marking To Do items as done

Test plan:
* Log in as student6 on narmstrong
* View To Do list and swipe to mark the item as done
* The item should not re-appear

### DIFF
--- a/CanvasCore/CanvasCore/Todo/Network/TodoAPI.swift
+++ b/CanvasCore/CanvasCore/Todo/Network/TodoAPI.swift
@@ -23,6 +23,6 @@ open class TodoAPI {
     }
 
     open class func ignoreTodo(_ session: Session, todo: Todo) throws -> URLRequest {
-        return try session.DELETE(todo.ignoreURL)
+        return try session.DELETE(URL(string: todo.ignoreURL)?.path ?? todo.ignoreURL)
     }
 }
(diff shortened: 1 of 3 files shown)


CanvasCore/CanvasCore/Todo/Network/TodoAPI.swift
```diff
@@ -23,6 +23,6 @@ open class TodoAPI {
     }
 
     open class func ignoreTodo(_ session: Session, todo: Todo) throws -> URLRequest {
-        return try session.DELETE(todo.ignoreURL)
+        return try session.DELETE(URL(string: todo.ignoreURL)?.path ?? todo.ignoreURL)
```
vanwagonet: why was this wrong before?

armstrongnate: `session.DELETE` and friends want paths, not full URLs. They append the passed in path to the `baseURL`, even if what's passed in is a fully qualified url.. 

I'm terrified of messing with Session and ThreeLegit stuff. This seemed like the least risky change.

I did attempt to make `Todo.ignoreURL` a `@NSManaged URL` but I guess `URI` types in core data models are a newish thing and the Todo data model file is too old.. I could bump up the version number but again, I'm afraid of breaking stuff.

vanwagonet: What if we still save a string, but it only be the path, since that's what we need.

inst-danger:
#### Release Note: 
Fixed marking To Do items as done

#### Affected Apps: Student

[MBL-13362](https://instructure.atlassian.net/browse/MBL-13362)

**Canvas iOS** | 93.25% | 93.25% | 0%
<p align="right">
</p>

PR #963, opened 2020-01-14
refactor RouteOptions to make RouterProtocol.show simpler

cobbal:
refs: none
affects: student, teacher, parent
release note: none

[run-nightly]


Core/Core/Router/Router.swift
```diff
@@ -18,115 +18,122 @@
...
         return route(to: .parse(url), from: from, options: options)
     }
 
-    func show(_ view: UIViewController, from: UIViewController, options: RouteOptions? = nil) {
+    func show(_ view: UIViewController, from: UIViewController, options: RouteOptions = .noOptions) {
         show(view, from: from, options: options, completion: nil)
     }
 
-    func show(_ view: UIViewController, from: UIViewController, options: RouteOptions? = nil, completion: (() -> Void)?) {
+    func show(_ view: UIViewController, from: UIViewController, options: RouteOptions = .noOptions, completion: (() -> Void)?) {
```
cobbal: Simplifying this function was the motivation behind this PR

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  You have included an invalid app. Valid values are: Student, Teacher, Parent, None
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


**Canvas iOS** | 93.45% | 93.77% | -0.31%
<p align="right">
</p>

Core/Core/Router/Router.swift
```diff
@@ -18,115 +18,122 @@
 
 import UIKit
 
-public struct RouteOptions: OptionSet {
-    public let rawValue: Int
+public struct RouteOptions: Equatable {
+    public let modal: Modal?
+    public let detail: Bool
+    public let embedInNav: Bool
+    public let addDoneButton: Bool
```
vanwagonet: so it seems like we have 3 different places the new view ends up, and some options that apply to each.
* `.push(addDoneButton: Bool = false)`
* `.detail(embedInNav: Bool = false, addDoneButton: Bool = false)`
* `.modal(_ style: UIModalPresentationStyle? = nil, isDismissable: Bool = true, embedInNav: Bool = false, addDoneButton: Bool = false)`

I thought about using an enum for this, but that may make getting the options back out more painful. So a struct is probably fine. Maybe an enum with `case push, detail, modal(style, dismissable)` as part of the struct would be fine. 🤷‍♂

Student/StudentUnitTests/People/PeopleListPresenterTests.swift
```diff
@@ -0,0 +1,138 @@
+//
```
vanwagonet: Zombie file. Please rebury.

Student/Student/People/PeopleListPresenter.swift
```diff
@@ -0,0 +1,69 @@
+//
```
vanwagonet: Zombie file. Please rebury.

Core/Core/Router/Router.swift
vanwagonet: maybe` push` shouldn't have `addDoneButton`, since it will have a back button. Any save button can't be trivially added by routing.

vanwagonet: maybe `detail` doesn't need an `addDoneButton` either. What would it do?

Core/Core/Router/Router.swift
```diff
@@ -18,43 +18,40 @@
...
-    public struct Modal: Equatable {
-        public let presentationStyle: UIModalPresentationStyle?
-        public let inPresentation: Bool
-    }
-
-    public init(
-        modal: Modal? = nil,
-        detail: Bool = false,
+public enum RouteOptions: Equatable {
+    case push(addDoneButton: Bool = false)
```
vanwagonet: I don't know if push needs an addDoneButton. I'm pretty sure we only do that for modals, but I didn't look at all of the uses to see if it does show up.

Core/Core/Router/Router.swift
```diff
@@ -18,43 +18,40 @@
...
+    case detail(embedInNav: Bool = false)
+    case modal(
+        _ style: UIModalPresentationStyle? = nil,
```
vanwagonet: I didn't realize swift lets you have defaults in enum cases, I thought we'd need a separate static func to handle default params. This is nice.

Core/Core/Router/Router.swift
```diff
@@ -18,43 +18,40 @@
...
 
-    public static let noOptions = RouteOptions()
+    public var embedInNav: Bool {
+        switch self {
+        case .detail(embedInNav: true), .modal(_, _, embedInNav: true, _):
+            return true
+        default:
+            return false
+        }
+    }
```
vanwagonet: these properties make it pretty straightforward to get options back out. nice.

Core/Core/Router/Router.swift
```diff
@@ -89,32 +86,33 @@ public extension RouterProtocol {
...
-            view.addDoneButton(side: .left)
-        }
         var nav: UINavigationController?
         if options.embedInNav {
             nav = view as? UINavigationController ?? UINavigationController(rootViewController: view)
         }
 
-        if let modal = options.modal {
+        switch options {
+        case let .modal(modalOptions):
```
vanwagonet: I thought the syntax was `case .modal(let style, let isDismissable, let embedInNav, let addDoneButton)`. Does what you did here just collect all the params into a tuple?

Core/Core/Router/Router.swift
```diff
@@ -89,32 +86,33 @@ public extension RouterProtocol {
...
-            if #available(iOS 13, *), modal.inPresentation {
+            if #available(iOS 13, *), !modalOptions.isDismissable {
                 (nav ?? view).isModalInPresentation = true
             }
             from.present(nav ?? view, animated: true, completion: completion)
-        } else if from.splitViewController != nil, options.detail, !from.isInSplitViewDetail {
+        case .detail where from.splitViewController != nil  && !from.isInSplitViewDetail:
             from.showDetailViewController(nav ?? view, sender: from)
-        } else {
+        default:
```
vanwagonet: I'd prefer to have `case .push:` here so if we added something else the compiler would force us to handle it.

cobbal: It's not quite a case .push though. It's also the .details where the "where" clause failed

Core/Core/Router/Router.swift
cobbal: It seems to, I didn't want to rely on the order of the variables, it would be easy to accidentally switch a `let isDismissable` with a `let embedInNav` with no warning in the above version.

Core/Core/Router/Router.swift
vanwagonet: case .detail, .push: then?

PR #1421, opened 2020-07-13
Add score statistics to assignment, display with graph

ndrewh:
This depends on this PR to add score statistics to the API: https://github.com/instructure/canvas-lms/pull/1669

Screenshot:

This feature is modeled after the (honestly, rather lame) web UI feature (it's a click-to-reveal thing on the student grades screen).
Screen Shot 2020-07-13 at 1 01 43 AM

What I implemented (the UI part, at least) is kinda overkill and I'd be perfectly happy if the app just displayed the min, max, and avg as text, but I thought I could do a little better.

Edge cases still display well:

    

Supports dynamic text sizing:

Tests included. Localization for the "min", "max", and "avg" strings is not included (if you even need that).

I would like to see this feature added to the app in some form (pending my API PR getting accepted), so if there's anything I can do to revise or rework this PR, let me know.

### DIFF
--- a/Core/Core/Assignments/Assignment.swift
+++ b/Core/Core/Assignments/Assignment.swift
@@ -55,6 +55,7 @@ public class Assignment: NSManagedObject {
     @NSManaged public var allDates: Set<AssignmentDate>
     @NSManaged public var allowedAttempts: Int // 0 is flag disabled, -1 is unlimited
     @NSManaged public var externalToolContentID: String?
+    @NSManaged public var scoreStatistics: ScoreStatistics?
 
     /**
      Use this property (vs. submissions) when you want the most recent submission
@@ -109,15 +110,15 @@ public class Assignment: NSManagedObject {
     public var isMasteryPathAssignment: Bool { masteryPathAssignment != nil }
 
     @discardableResult
-    public static func save(_ item: APIAssignment, in context: NSManagedObjectContext, updateSubmission: Bool) -> Assignment {
+    public static func save(_ item: APIAssignment, in context: NSManagedObjectContext, updateSubmission: Bool, updateScoreStatistics: Bool) -> Assignment {
         let assignment: Assignment = context.first(where: #keyPath(Assignment.id), equals: item.id.value) ?? context.insert()
-        assignment.update(fromApiModel: item, in: context, updateSubmission: updateSubmission)
+        assignment.update(fromApiModel: item, in: context, updateSubmission: updateSubmission, updateScoreStatistics: updateScoreStatistics)
         return assignment
     }
 }
 
 extension Assignment {
-    func update(fromApiModel item: APIAssignment, in client: NSManagedObjectContext, updateSubmission: Bool) {
+    func update(fromApiModel item: APIAssignment, in client: NSManagedObjectContext, updateSubmission: Bool, updateScoreStatistics: Bool) {
         id = item.id.value
         name = item.name
         courseID = item.course_id.value
@@ -183,6 +184,17 @@ extension Assignment {
             } else if let submissions = submissions {
                 client.delete(Array(submissions))
                 self.submissions = nil
+            }
+        }
+        
+        if updateScoreStatistics {
+            if let newStatistics = item.score_statistics {
+                let replacementStats = scoreStatistics ?? client.insert()
+                replacementStats.update(fromApiModel: newStatistics, in: client)
+                self.scoreStatistics = replacementStats
+            } else if let scoreStatistics = scoreStatistics {
+                client.delete(scoreStatistics)
+                self.scoreStatistics = nil
             }
         }
 
@@ -315,5 +327,19 @@ public final class AssignmentDate: NSManagedObject {
         model.unlockAt = item.unlock_at
         model.lockAt = item.lock_at
         return model
+    }
+}
+
+public final class ScoreStatistics: NSManagedObject {
+    @NSManaged internal (set) public var mean: Double
+    @NSManaged internal (set) public var min: Double
+    @NSManaged internal (set) public var max: Double
+    @NSManaged internal (set) public var assignment: Assignment
+    
+    
+    public func update(fromApiModel item: APIAssignmentScoreStatistics, in client: NSManagedObjectContext) {
+        mean = item.mean
+        min = item.min
+        max = item.max
     }
 }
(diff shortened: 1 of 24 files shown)


ndrewh:
One thing I was thinking: this may be more appropriate if was hidden inside some collapsing view (or in some other way was hidden by default, like it is in the web ui). I didn't find anywhere else in your UI that you tap on something and it 'expands' like the web ui does, but this could be implemented as some kind of modal or popup from the bottom, or something like that if that is desired. (in my opinion it's fine where it is, but just offering other options :)

Also some notes:
- My changes include adding support for score statistics to CoreCanvas, but I didn't use it or test it there (I found out it wasn't used in the parts of the app I cared about). If you want me to remove the CoreCanvas parts I can do so (or feel free to yourself). The additions I made to Core, on the other hand, _do_ include tests.
- I can't see the failing tests on bitrise, so if there are legitimately failing tests, let me know. I can run most of them (including all the ones I wrote) on my machine and I was finally able to get the UI tests to build, so I'll be fixing those momentarily. There's one of the login tests that intermittently crashes my simulator, but I'm going to assume that's unrelated.

armstrongnate:
Hi @ndrewh and thank you for the contribution. This is a cool thing to add to mobile. In order for us to ship it we will need to add it on android as well. Have you thought about adding it there too? https://github.com/instructure/canvas-android

If not, we'll have to wait until we can get it on the product roadmap for android. In the meantime you will need to put it behind a feature flag (see `ExperimentalFeature.swift`) so that we can still release from `master`. The feature flag will also allow us to merge without the API changes in canvas-lms.

I will attempt to get both of your PRs running locally so that I can test it out.

CanvasCore/CanvasCore/Assignments/Assignment/Assignment.swift
```diff
@@ -75,7 +75,8 @@ public final class Assignment: NSManagedObject, LockableModel {
 
     @NSManaged internal (set) public var rubric: Rubric?
     @NSManaged internal (set) public var assignmentGroup: AssignmentGroup?
-
+    @NSManaged internal (set) public var scoreStatistics: ScoreStatistics?
```
armstrongnate, line 78: Yeah, we won't need these changes in CanvasCore.

Core/Core/Assignments/Assignment.swift
```diff
@@ -183,6 +184,21 @@ extension Assignment {
...
+        
+        if updateScoreStatistics {
+            if let newStatistics = item.score_statistics {
+                if let scoreStatistics = scoreStatistics {
+                    scoreStatistics.update(fromApiModel: newStatistics, in: client)
+                } else {
+                    let saveStats = ScoreStatistics(context: client)
+                    saveStats.update(fromApiModel: newStatistics, in: client)
+                    scoreStatistics = saveStats
+                }
```
armstrongnate, line 196: This `if/else` could be simplified to:

```swift
scoreStatistics = scoreStatistics ?? client.insert()
scoreStatistics.update(fromApiModel: newStatistics, in: client)
```

Core/Core/Views/GradeStatisticGraphView.swift
```diff
@@ -0,0 +1,140 @@
...
+        setupGraph(assignment: assignment)
+    }
+    
+    func setupGraph(assignment: Assignment) {
+        guard let stats = assignment.scoreStatistics, let points_possible = assignment.pointsPossible, let score = assignment.viewableScore, points_possible > 0 else {
+            isHidden = true
+            return
+        }
+        isHidden = false
+        let usableWidth = graphArea.frame.width - 48.0 - 2.0 // subtract 2 for half width of the outermost bars
```
armstrongnate, line 73: I haven't tried it yet but I suspect that this won't handle window size changes like when the user changes the width of the app when in side-by-side mode on the iPad. You might have to move the offset calculations and the constraint-constant-adjustments to `viewDidLayoutSubviews`. I could be wrong. It's possible `update(_:)` gets called at all the necessary times. I will try it out and see.

Core/Core/Views/GradeStatisticGraphView.swift
```diff
@@ -0,0 +1,140 @@
...
+        }
+        
+        // We want the layout to update NOW
+        layoutIfNeeded()
+
+    }
+}
+
+extension ClosedRange {
+    func clamp(_ value : Bound) -> Bound {
```
armstrongnate, line 135: You could move this to a new file `Core/Extensions/ClosedRangeExtensions.swift`

ndrewh:
I'll work on modifying to check for an experimental feature flag tonight.

I'll look into Android too -- does the graph have to match the iOS app very closely? I might be able to implement the graph on Android but I probably won't be able to do anything as fancy as what I did here with the text labels aligning with the graph... will that be good enough? Like I could still do the graph but then above it would be statically-placed (or in a stack view or something) text labels for min, max, mode. Just wondering if the android app has to be identical or anything.

Thanks!

armstrongnate:
@ndrewh for android I would recommend getting as close as you can and we'll go from there. It doesn't have to be perfect. And we can help polish if needed.

ndrewh:
@armstrongnate Here's a mostly-working prototype in android: https://github.com/instructure/canvas-android/pull/888

I haven't added the feature test yet, but I will do that soon-ish

armstrongnate:
Hey @ndrewh we want to keep the design on both platforms the same and we like how you have it on Android with the labels statically placed. Can you do that on iOS as well?

armstrongnate:
And just so you know, product has approved this feature so we will wait for the feature flag and the other requested changes (or the canvas-lms change + android changes, whichever comes first) before pulling it down and reviewing it further.

Core/Core/Assignments/Assignment.swift
ndrewh: Turns out this doesn't actually work: because scoreStatistics is an optional, when insert() tries to convert the type to a string, it gets "Optional<ScoreStatistics>" which is not a valid name of a model. I will still refactor it a bit.

Core/Core/Views/GradeStatisticGraphView.swift
ndrewh: I switched this to UIView's layoutSubviews (prior to calling `super.layoutSubviews()`) and the result is great. I even get this sweet resizing animation, with zero additional code:

ezgif-3-f368d05a9ad2

ndrewh:
@armstrongnate I made all of the requested changes. 



I added the feature flag in a bunch of places -- there were some I was unsure if they needed it or not. Also, feel free to change the font weight or size for the min/max/avg labels.

ndrewh:
@armstrongnate Anything I need to do here? The canvas-lms change seems to be stalled.

ndrewh:
The canvas-lms change got merged. I don't know what the release schedule is for canvas-lms so I don't know when this will go live.

I've removed the feature flag from this PR and rebased. It should be ready for review @armstrongnate 

Thanks,
@ndrewh

ndrewh:
Looks like the api change has hit beta canvas-lms instances. Is there anything for me to do here? @armstrongnate

armstrongnate:
Hi @ndrewh, I will be looking at this and the android change today. Thanks for your patience.

armstrongnate:
Moving to #1519

PR #1496, opened 2020-08-21
Finish Parent QR account creation

armstrongnate:
refs: MBL-14413
affects: parent, student
release note: Added pairing with students from the launch screen

[run-nightly]

Test plan:
- Log into Parent and enable qrLoginParent and parentQRCodePairing
- Log out of Parent
- Tap the QR button
- Tap 'I don't have a Canvas account'
- Download Student app and enable `studentQRCodePairing`
- Settings -> Pair with Observer
- Scan the QR code from the Parent app
- You should be taken to the Create Account screen with the pairing code set
- Create a new account and tap Start Participating
- You should be logged in and paired with the user
- Log in as the same parent on web
- Log out of the Parent app
- Tap the QR button
- Tap 'I have a Canvas account'
- Scan the QR code from canvas web to login
- You should be logged in
- Tap add Student from the Parent dashboard
- Use the QR scanner to add a second child


Parent/ParentE2ETests/LoginTests.swift
```diff
@@ -0,0 +1,36 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Core
+import TestsFoundation
+
+class LoginCreateAccountE2ETests: CoreUITestCase {
+    override var experimentalFeatures: [ExperimentalFeature] { [.qrLoginParent, .parentQRCodePairing] }
+    override var qrCode: String? { "canvas-parent://iosauto.beta.instructure.com/pair?code=abc" }
```
cobbal, line 24: I don't like that this only works if "launch" is called. If this is the only place it's used, something like this might be better:

```suggestion
    override func setup() {
        launch { app in 
            app.launchEnvironment["QR_CODE"] = "canvas-parent://iosauto.beta.instructure.com/pair?code=abc"
        }
        super.setup()
    }
```

armstrongnate: True, not safe to assume launch will get called. I'll make this change and remove the code in the super class as well.

PR #1580, opened 2020-10-08
Configured app to bind to 8082 instead of 8081 for local dev

kylehovey:
Port 8081 is used by the canvas-dev vagrant setup. It is easier to change the port here than it would be for the vagrant setup.


kylehovey:
This was originally created for a forked repository and automatically targeted the `instructure/canvas-ios` repo. Sorry about that.

PR #1944, opened 2021-08-26
Disable pairing if self registration is disallowed

vargaat:
refs: MBL-15532
affects: Student
release note: none

test plan: See ticket. Pairing option in settings should only be visible if self registration is set to all or observer on the canvas account.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>


idevainstr (approved):
QA+1

balintbartok (approved):
QA + 1

PR #1993, opened 2021-10-27
Fix SpeedGrader comment field content disappearing and other layout issues

vargaat:
refs: MBL-15689
affects: Teacher
release note: Fixed entered submission comment disappearing upon rotation on iPads.

test plan:
- On iPad go to SpeedGrader.
- Enter a comment into the text field but don't send it, keep the keyboard on screen.
- Put the app into background then move it to foreground.
- Entered text shouldn't disappear.
- Repeat this for the other interface orientation as well.
---
- On iPad go to SpeedGrader.
- Enter a comment into the text field but don't send it, keep the keyboard on screen..
- Rotate the device to the other orientation.
- Entered text shouldn't disappear.
- Keyboard should disappear, its are should be filled back with our content.
---
- On iPad go to SpeedGrader on a submission which can be annotated.
- Put the app into background then move it to foreground.
- Submission content shouldn't reload.
---
- iPhone SpeedGrader should work as before.
---
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>


balintbartok (approved):

davidvarga93 (approved):
QA + 1

idevainstr (approved):
QA + 1

PR #2137, opened 2022-04-12
Course cards show grades as 0% when navigating from People page to Dashboard

idevainstr:
refs: MBL-15961
affects: Student
release note: Fixed grade displaying 0% when navigation back from the People page.
test plan: see ticket

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

### DIFF
--- a/Core/Core/Enrollments/Enrollment.swift
+++ b/Core/Core/Enrollments/Enrollment.swift
@@ -135,7 +135,7 @@ extension Enrollment {
 
         self.course = course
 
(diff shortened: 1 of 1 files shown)


vargaat (changes requested):
It shows N/A for me instead of 0%, see the video I attached to the ticket.

idevainstr:
Shelved.
As per out current database handling there's no easy fix for this. The dashboard has a forced refresh (and also pull to refresh) at the moment that refetches the grade, this should be sufficient as a workaround for now.

PR #2272, opened 2022-10-03
Use waitUntil instead of running the main RunLoop for a fixed amount of time in unit tests.

vargaat:
[ignore-commit-lint]


inst-danger:
**Canvas iOS** | 91.69% | 91.68% | 0.01%
<p align="right">
</p>

szabolcs-instructure (approved):

idevainstr (approved):
Shouldn't we preserve the wait times?

vargaat:
In these tests those wait times were there just to ensure some async code had enough time to run in parallel of the test. Those wait times were not reliable as sometimes it needed even more that what we've had there but most of the times it just executed immediately. With this wait method it polls the state every 0.1 secs and completes when it reached the desired condition. So the test most of the time will finish quicker while also allows slower machines to successfully run the test.

balintbartok (approved):

PR #2286, opened 2022-10-19
Improve mobile login

idevainstr:
refs: MBL-14653
release note: Improved the login flow by saving the last institution's login page.
test plan: See ticket.

## Checklist

- [x] Follow-up e2e test ticket created or not needed
- [x] A11y checked
- [x] Tested on phone
- [x] Tested on tablet
- [x] Approve from product or not needed


inst-danger:

</details>

inst-danger:

</details>

inst-danger:

</details>

Core/Core/Login/LoginFindSchoolViewController.swift
```diff
@@ -156,6 +157,11 @@ class LoginFindSchoolViewController: UIViewController {
             navigationItem.rightBarButtonItem = nil
         }
     }
+
+    func saveAccount(_ account: APIAccountResult?) {
```
vargaat, line 161: It seems that we only use the `domain` field of this `APIAccountResult`. Would it be possible to save this `String` instead of the whole object?

vargaat (changes requested):
Can we save the domain only after a successful login? If I enter an invalid domain and don't log in, the app will still offer me my last entered invalid link.

Core/Core/Login/LoginFindSchoolViewController.swift
balintbartok: If I remember correctly, the authentication method was also useful, but it's not used in this solution, is there a reason for that?

balintbartok: also, the "name" field is used

idevainstr: I used the object on purpose.

szabolcs-instructure (approved):
QA + 1

Core/Core/Login/LoginFindSchoolViewController.swift
vargaat: What was that purpose?

idevainstr: To access several variable. We use name and domain.

vargaat: Ok, I missed the case when an account is selected from the previous logins list.

Core/Core/Login/LoginFindSchoolViewController.swift
```diff
@@ -146,6 +146,7 @@ class LoginFindSchoolViewController: UIViewController {
             host = "\(host).instructure.com"
         }
         searchField.resignFirstResponder()
+        env.lastLoginAccount = APIAccountResult(name: "", domain: host, authentication_provider: nil)
```
vargaat, line 149: Is this doing anything? If I don't log in just go back from the login screen to the welcome page I don't see any change. Should we update this `lastLoginAccount` field at all prior to logging in?

vargaat (changes requested):
Starting the app without any accounts displays both buttons but both lead to the find my school screen.

Core/Core/Login/LoginFindSchoolViewController.swift
idevainstr: We need to pass the selected account, then save it after login.

Core/Core/Login/LoginStartViewController.swift
```diff
@@ -78,10 +86,16 @@ class LoginStartViewController: UIViewController {
...
         wordmarkLabel.textColor = .currentLogoColor()
-
+        let loginText = NSLocalizedString("Log In", bundle: .core, comment: "")
         if MDMManager.shared.host != nil {
-            findSchoolButton.setTitle(NSLocalizedString("Log In", bundle: .core, comment: ""), for: .normal)
+            findSchoolButton.setTitle(loginText, for: .normal)
+        } else if let data = UserDefaults.standard.data(forKey: "lastLoginAccount"),
+                    let savedAccount = try? APIJSONDecoder().decode(APIAccountResult.self, from: data) {
+            let buttonTitle = savedAccount.name.isEmpty ? savedAccount.domain : savedAccount.name
+            findSchoolButton.setTitle(NSLocalizedString(buttonTitle, bundle: .core, comment: ""), for: .normal)
```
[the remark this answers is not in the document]
idevainstr: Good idea, moved it.

szabolcs-instructure:
QA + 1

vargaat (approved):
QA+1 We should think about making this available for QR logins as well.

szabolcs-instructure (approved):

davidvarga93 (dismissed):
QA + 1

vargaat (approved):
QA+1 Login buttons now don't fill the whole screen on iPad.

PR #2480, opened 2023-06-12
Update developer menu

vargaat:
refs: MBL-16706
affects: Student, Teacher
release note: none

test plan:
- Open developer menu.
- Design should be according to the latest app standards.
- Cells should register taps when touched on an empty area.
- Open a menu item and close it.
- Menu items shouldn't duplicate.
- Two new items should be available. One for the app directory and another for the shared container dir.


szabinst (approved):
QA + 1

idevainstr (approved):
QA +1

ndrsszsz (approved):
QA +1

PR #2553, opened 2023-08-24
Offline sync retry logic

szabinst:
refs: MBL-16885
affects: Student
release note: none
test: 
1. I suggest to use my account because "Test" course has 100+ MB of files so you can play around with the retry logic. 
2. Start a course sync 
3. Open the progress screen
4. Turn off Wi-Fi
5. Turn back the Wi-Fi (To have an established connection you might need to wait an additional second after the Wi-Fi icon appears) 
6. Tap Retry 
7. You should see the previously failed Course/Tab/File row switches state from Error to Loading
8. If you previously successfully downloaded a couple of files, the total number of bytes should decrease because we are only downloading the ones that failed.
 
https://github.com/instructure/canvas-ios/assets/110813321/71f9f931-87a0-45a4-b250-a8ee06ceaa01

## Checklist

- [ ] Tested on phone
- [x] Tested on tablet

### DIFF
--- a/Core/Core/CourseSync/CourseSyncProgress/ViewModel/CourseSyncProgressInfoViewModel.swift
+++ b/Core/Core/CourseSync/CourseSyncProgress/ViewModel/CourseSyncProgressInfoViewModel.swift
@@ -48,6 +48,7 @@ class CourseSyncProgressInfoViewModel: ObservableObject {
 
         interactor.observeDownloadProgress()
             .receive(on: scheduler)
+            .throttle(for: .milliseconds(300), scheduler: scheduler, latest: true)
             .sink { state in
                 switch state {
                 case .data(let progressList):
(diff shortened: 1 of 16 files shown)


inst-danger:

</details>

balintbartok (dismissed):
QA + 1

Core/Core/CourseSync/CourseSyncProgress/ViewModel/CourseSyncProgressInfoViewModel.swift
```diff
@@ -48,25 +48,27 @@ class CourseSyncProgressInfoViewModel: ObservableObject {
 
         interactor.observeDownloadProgress()
             .receive(on: scheduler)
+            .throttle(for: .milliseconds(300), scheduler: scheduler, latest: true)
```
balintbartok, line 51: I know it's not the first time this is introduced, but what's the reason behind the 300 ms? why 300?

szabinst: I removed both throttles.

vargaat (changes requested):
I think the sync failed states should be reset on all items after retry is pressed. Currently only the first batch of files states are updated.

https://github.com/instructure/canvas-ios/assets/72396990/7cef0ff8-995d-4151-9142-fa2995d0a07d

szabinst:
@vargaat Made the changes, please check again.

vargaat (approved):
QA+1

The progress bar flickers like mad upon retry but we can improve this UI glitch later.

idevainstr (approved):
QA +1

PR #2561, opened 2023-08-29
Offline course sync list interactor

szabinst:
Extracted common logic into a separated interactor that is capable of listing courses based on a filter. 

```
    enum Filter {
        case courseID(String) // A specific course
        case all // All the available courses
        case synced // Synchronized courses
    }
```

Case `courseID` and `all` is used on the selector screen, while `synced` is used on the progress screen and will be used for automated background downloads in the future. 

refs: MBL-16887
affects: Student
release note: none
test plan: Test if selection and progress screen works as before

## Checklist

- [x] Tested on phone
- [ ] Tested on tablet


idevainstr (approved):
QA +1

Core/Core/CourseSync/Common/Models/CourseSyncListerInteractor.swift
```diff
@@ -0,0 +1,177 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Combine
+import CombineSchedulers
+import Foundation
+
+protocol CourseSyncListInteractor {
+    func getCourseSyncEntries(filter: CourseSyncListInteractorLive.Filter) -> AnyPublisher<[CourseSyncEntry], Error>
```
vargaat, line 24: I used to extract such enums into a separate entity. It doesn't seem to be correct that a protocol definition uses a type from a concrete implementation of the protocol.

vargaat (changes requested):
QA+1

Core/Core/CourseSync/Common/Models/CourseSyncListerInteractor.swift
szabinst: I agree, was a bit hesitant but ended up leaving it there. I moved it to a separate entity and a separate file.

vargaat: It would be nice if we could embed an enum into a protocol.

vargaat (approved):

vargaat (changes requested):
There's a compilation error in the test target.

idevainstr (approved):

PR #2711, opened 2024-01-31
Disable Discussions tab in offline mode if discussionRedesign is enabled

szabinst:
Disable Discussions tab in offline mode if discussionRedesign is enabled.

affects: Student

[ignore-commit-lint]

## Checklist

- [ ] Follow-up e2e test ticket created
- [ ] Tested on phone
- [ ] Tested on tablet

PR #2713, opened 2024-02-08
Student - E2E test for the new discussion screen

ndrsszsz:
refs: MBL-17336
affects: None
release note: None
test plan: Tests to pass

### E2E test case for the new discussion screen
1.  Turns on the feature flag with API call
2. Creates a user, course, discussion with DataSeeder
3. Gets the user logged in
4. Navigates to Discussions screen and taps on the discussion
5. Checks elements on the new discussion details screen
6. Turns off the feature flag (in teardown function)

### DIFF
--- a/Student/StudentE2ETests/Discussions/DiscussionsTest.swift
+++ b/Student/StudentE2ETests/Discussions/DiscussionsTest.swift
@@ -22,6 +22,11 @@ class DiscussionsTests: E2ETestCase {
     typealias Helper = DiscussionsHelper
     typealias DetailsHelper = Helper.Details
     typealias ReplyHelper = DetailsHelper.Reply
+    typealias NewDiscussion = Helper.NewDetails
+
+    override func tearDown() {
+        seeder.setFeatureFlag(featureFlag: .newDiscussion, state: .off)
+    }
 
     func testDiscussionLabels() {
         // MARK: Seed the usual stuff with a discussion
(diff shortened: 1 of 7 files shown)


inst-danger:

</details>

inst-danger:
#### Release Note: 
None

#### Affected Apps: None

[MBL-17336](https://instructure.atlassian.net/browse/MBL-17336)

**Canvas iOS** | 92.03% | 92.03% | 0%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7470e34de0410536fc2ad9ca15cc678e84b193ff
</p>

szabinst (changes requested):
The discussion test looks good. 

Please add a test that covers Announcements as well since Announcements are technically Discussions. 

We could add an other test that covers [this](https://github.com/instructure/canvas-ios/pull/2711) use case.

ndrsszsz:
> The discussion test looks good.
> 
> Please add a test that covers Announcements as well since Announcements are technically Discussions.
> 
> We could add an other test that covers [this](https://github.com/instructure/canvas-ios/pull/2711) use case.

@szabinst Thanks for the suggestions, all is done! Let me know if I still missed something. 🙂

szabinst (changes requested):
The newly added tests look good, but now we have some Xcode warnings. 


Screenshot 2024-02-13 at 13 14 28

Student/StudentE2ETests/Discussions/DiscussionsTest.swift
```diff
@@ -22,6 +22,12 @@ class DiscussionsTests: E2ETestCase {
     typealias Helper = DiscussionsHelper
     typealias DetailsHelper = Helper.Details
     typealias ReplyHelper = DetailsHelper.Reply
+    typealias NewDiscussion = Helper.NewDetails
+
+    override func tearDown() {
+        let featureFlagResponse = seeder.setFeatureFlag(featureFlag: .newDiscussion, state: .off)
+        XCTAssertEqual(featureFlagResponse.state, DSFeatureFlagState.off.rawValue)
```
vargaat, line 29: `super` isn't called after we tore down our state. Was this on purpose?

Student/StudentE2ETests/Offline/FeatureFlagOfflineTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+        super.setUp()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+
+        // Disable discussion redesign feature flag
+        let featureFlagResponse = seeder.setFeatureFlag(featureFlag: .newDiscussion, state: .off)
+        XCTAssertEqual(featureFlagResponse.state, DSFeatureFlagState.off.rawValue)
```
vargaat, line 35: The `super` call order order should be inverted.
```
super.setup()
subclass.setup()
subclass.tearDown()
super.tearDown()
```

Student/StudentE2ETests/Announcements/AnnouncementsTests.swift
```diff
@@ -23,6 +23,14 @@ class AnnouncementsTests: E2ETestCase {
     typealias DetailsHelper = Helper.Details
     typealias AccountNotifications = Helper.AccountNotifications
 
+    override func tearDown() {
+        super.tearDown()
+
+        // Disable discussion redesign feature flag
+        let featureFlagResponse = seeder.setFeatureFlag(featureFlag: .newDiscussion, state: .off)
+        XCTAssertEqual(featureFlagResponse.state, DSFeatureFlagState.off.rawValue)
```
vargaat, line 31: `super.tearDown()` should be the last to call in this override.

vargaat (dismissed):

Student/StudentE2ETests/Discussions/DiscussionsTest.swift
ndrsszsz: Nice catch, thanks! 👍

Student/StudentE2ETests/Offline/FeatureFlagOfflineTests.swift
ndrsszsz: This is intentional here. This is an offline test, which means that the network connection gets switched back in the superclass teardown call, while here in this teardown call a request gets sent (which is not possible if the network connection is still off) so this is why I called the teardown of the superclass first.

Student/StudentE2ETests/Announcements/AnnouncementsTests.swift
ndrsszsz: Correct, done.

Student/StudentE2ETests/Offline/FeatureFlagOfflineTests.swift
ndrsszsz: Done anyway. 🙂

vargaat (approved):

PR #2776, opened 2024-04-24
Large GIF course image causes app to crash

rh12:
refs: [MBL-17453](https://instructure.atlassian.net/browse/MBL-17453)
affects: Teacher, Student
release note: Optimized memory and bandwidth usage for Course Image and uploaded files.

## Problem
- Animated GIFs were loaded as bitmaps frame by frame, which eats up a lot of memory.
- Image caching was based on the full image URLs, which included token in the query as well. Because it changes slighly in each request, this resulted in each Course Image downloaded and loaded into memory multiple times.
- In-memory image cache was never reseted.

## What's new
- Animated GIFs are displayed and animating in WebViews, where deemed important:
  - Course Details screen
  - Customize Course screen
  - Course Settings screen
  - Files
  - K5 Subject Details screen
- Animated GIFs are showing only the first frame, without animating in all other places, including:
  - Dashboard
  - K5 Homeroom screen
- Image caching now ignores token

## What's not changed
- Animated GIFs may still be downloaded multiple times, but this is no longer a noticeable issue.
- In-memory image cache is still not reseted, because it is hard to guess when it's no longer needed.

## Test plan
See ticket and use already setup account described there.
- Verify GIFs are animating and apps doesn't crash in places listed above.
- Verify non-animated Course Images load instantly in Course Details screen if it was already loaded & displayed on Dashboard.
- Verify Animated GIFs are zoomable and scrollable in Files.
- Verify Animated GIFs are not zoomable or scrollable in all other places.

## Checklist

- [ ] A11y checked
- [x] Tested on phone
- [x] Tested on tablet
- [ ] Tested in dark mode
- [x] Tested in light mode
- [ ] Approve from product

[MBL-17453]: https://instructure.atlassian.net/browse/MBL-17453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Courses/CourseDetails/View/CourseDetailsHeaderView.swift
+++ b/Core/Core/Courses/CourseDetails/View/CourseDetailsHeaderView.swift
@@ -31,7 +31,7 @@ struct CourseDetailsHeaderView: View {
         ZStack {
             Color(viewModel.courseColor.resolvedColor(with: .light).darkenToEnsureContrast(against: .textLightest)).frame(width: width, height: viewModel.height)
             if let url = viewModel.imageURL {
-                RemoteImage(url, width: width, height: viewModel.height)
+                RemoteImage(url, width: width, height: viewModel.height, shouldHandleAnimatedGif: true)
                     .opacity(viewModel.imageOpacity)
                     .accessibility(hidden: true)
             }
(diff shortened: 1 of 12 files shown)


inst-danger:

</details>

vargaat (approved):
QA+1, Good job getting this together!

szabinst (approved):
QA + 1

The only thing that I'd change is the release note. "Optimized memory and bandwidth usage" is too technical in my opinion. Instead we could say something like "Improved performance on displaying large size GIF images across the app."

rh12:
> QA + 1
> 
> The only thing that I'd change is the release note. "Optimized memory and bandwidth usage" is too technical in my opinion. Instead we could say something like "Improved performance on displaying large size GIF images across the app."

Good point, thank you! Although I would drop the "GIF", as it helps with all images.

PR #2802, opened 2024-05-30
Skip offline sync without courses

szabinst:
Skip offline sync without selected courses.

refs: MBL-17204
affects: Student
release note: Offline sync skips cases when there are no courses selected
test plan: Build from Xcode to easily trigger a sync, check if sync only occurs when there are offline selected courses.

## Checklist

- [x] Tested on phone
- [ ] Tested on tablet


inst-danger:

</details>

vargaat (changes requested):
I accidentally was able to reproduce the "0 courses have been synced notification." the following way:
- I had a course that was selected for offline sync in the past.
- This course had course participation setup with and end date and the "Restrict students from viewing course after course end date" option was turned on.
- The end date has passed and the course disappeared both from the dashboard and from the all courses screen.
- The auto sync still started since there was a course selected to be synced.
- I received the 0 courses synced notification.

I'm not sure yet that at which point should we clear this course or if this can be avoided at all if the user doesn't open the offline sync selector screen after a course has disappeared.

szabinst:
@vargaat That's a really nice catch, I'll think about it.

rh12 (approved):
QA + 1

vargaat (approved):

PR #2827, opened 2024-06-26
Bump minimum OS version to iOS 16

szabinst:
refs: MBL-17070
affects: Student, Teacher, Parent
release note: The minimum supported OS version has been increased to iOS 16
test plan: Test everything

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [x] Tested on tablet
- [x] Tested in dark mode
- [ ] Tested in light mode

PR #2832, opened 2024-07-04
Calendar Edit ToDo

rh12:
refs: [MBL-17290](https://instructure.atlassian.net/browse/MBL-17290)
affects: Student
release note: Calendar To Dos are now editable.

## What's changed
- Added Edit ToDo screen
- Added kebab button for it to ToDo Details screen, which opens it directly for now
- Updated ToDo Details screen navbar to match calendar color
- Added "Title" label before Create ToDo screen's Title TextField
- Fixed Create ToDo error handling to fail for error message in body (i.e.: real backend errors)
  - it fails because decoding fails
  - it still won't fail if only HTTP status code is changed, but the body is valid
  - related Slack [conversation](https://instructure.slack.com/archives/G01GBRGPKQC/p1718898982585849?thread_ts=1718893358.967919&cid=G01GBRGPKQC)
- Save button is enabled only after user made any changes (if title is not empty)
  - it doesn't compare values, just observes a dirty flag

### Chores
- Added `View.errorAlert()` (similar to `View.confirmationAlert()`)
- Added convenience accessors and init to `Context`
- Added to `InstUI.NavigationBarButton`
  - `isBackgroundContextColor` parameter
  - image init
  - offline observation
- Added `OfflineObservingButton` (it's based on `PrimaryButton`)

## Test plan
- Verify Create ToDo screen still works as expected
- Verify Edit ToDo screen works as expected
- See above for details

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist
- [x] Follow-up e2e test ticket created
- [ ] A11y checked
- [x] Tested on phone
- [ ] Tested on tablet
- [x] Tested in dark mode
- [x] Tested in light mode
- [ ] Approve from product

[MBL-17290]: https://instructure.atlassian.net/browse/MBL-17290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


inst-danger:

</details>

inst-danger:
#### Release Note: 
Calendar To Dos are now editable.

#### Affected Apps: Student

[MBL-17290](https://instructure.atlassian.net/browse/MBL-17290)

**Canvas iOS** | 91.47% | 91.44% | 0.03%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0f49a4f4406ac8fc331b9547a821ae8aff472016
</p>

Core/Core/InstUI/Views/Cells/TextFieldCell.swift
```diff
@@ -99,14 +98,25 @@ extension InstUI {
...
         private var textField: some View {
-            TextField("", text: $text, prompt: Text(placeholder).foregroundColor(Color.placeholderGray))
-                .focused($isFocused)
-                .multilineTextAlignment(label == nil ? .leading : .trailing)
-                .font(label == nil ? .semibold16 : .regular16, lineHeight: .fit)
-                .foregroundStyle(Color.textDarkest)
-                .submitLabel(.done)
-                .accessibilityLabel(accessibilityLabel)
-                .accessibilityValue(accessibilityValue)
+            if #available(iOS 16, *) {
```
rh12, line 101: To be updated once [iOS16 PR](https://github.com/instructure/canvas-ios/pull/2827) is merged

Core/Core/Planner/CalendarToDo/ViewModel/CalendarToDoDetailsViewModel.swift
```diff
@@ -17,17 +17,40 @@
...
+        self.plannable = plannable
+
+        interactor.getToDo(id: plannable.id)
+            .sink { [weak self] in
+                self?.title = $0.title
+                self?.date = $0.date?.dateTimeString
+                self?.description = $0.details
+                self?.navBarColor = $0.color.ensureContrast(against: .backgroundLightest)
+            }
+            .store(in: &subscriptions)
```
szabinst, line 44: I see that this function fetches data from the db only. How come we swallow every error? If something fails during the Core Data fetch we'll have an empty white screen.

Core/Core/Planner/CalendarToDo/ViewModel/EditCalendarToDoViewModel.swift
```diff
@@ -49,46 +85,79 @@ final class EditCalendarToDoViewModel: ObservableObject {
...
+            mode = .edit(id: plannable.id)
+        } else {
+            mode = .add
+        }
+
+        title = plannable?.title ?? ""
+        date = plannable?.date ?? Clock.now.endOfDay() // end of today, to match default web behaviour
+        details = plannable?.details ?? ""
+
+        subscribeIsTouched(to: $title)
```
szabinst, line 121: I'm not sure if there's a better name for this logic instead of `isTouched` but I didn't get its meaning for the first time

Core/Core/Planner/CalendarToDo/ViewModel/EditCalendarToDoViewModel.swift
```diff
@@ -49,46 +85,79 @@ final class EditCalendarToDoViewModel: ObservableObject {
...
         toDoInteractor: CalendarToDoInteractor,
         calendarListProviderInteractor: CalendarFilterInteractor,
         completion: @escaping (PlannerAssembly.Completion) -> Void
     ) {
         self.toDoInteractor = toDoInteractor
         self.calendarListProviderInteractor = calendarListProviderInteractor
 
-        // end of today, to match default web behaviour
-        date = .now.endOfDay()
+        if let plannable {
```
szabinst, line 111: You could create a local `weak var weakSelf = self` property above this line so you wouldn't need to capture self in every closure.

Core/Core/Planner/CalendarToDo/ViewModel/EditCalendarToDoViewModel.swift
```diff
@@ -117,10 +179,37 @@ final class EditCalendarToDoViewModel: ObservableObject {
...
-            .sink()
+    private func subscribeIsTouched<T: Equatable>(to publisher: Published<T>.Publisher) {
+        publisher
+            .removeDuplicates()
+            .dropFirst()
+            .sink { [weak self] _ in self?.isTouched = true }
             .store(in: &subscriptions)
     }
+
+    private func saveAction() -> AnyPublisher<Result<Void, Error>, Never> {
```
szabinst, line 192: This return type seems a bit more complex than it should be. What about `AnyPulisher<Void, Error>` ?

szabinst (changes requested):
I'm not sure how other errors (like non 200 http codes) are handled but without network connection there's no error handling. 

Following other areas of the app, create and the edit button should not be enabled when offline mode is enabled. 

https://github.com/instructure/canvas-ios/assets/110813321/667b86c0-0e68-4230-924a-2e7287d98762

Core/Core/Planner/CalendarToDo/ViewModel/EditCalendarToDoViewModel.swift
rh12: It's reworked without `mapToResult()`

Core/Core/Planner/CalendarToDo/ViewModel/EditCalendarToDoViewModel.swift
rh12: updated to `isFieldsTouched`, as discussed

Core/Core/Planner/CalendarToDo/ViewModel/EditCalendarToDoViewModel.swift
rh12: I simplified these a bit, but I would rather keep the separate capturings.
My main reason is syntax highlight, which shows `self?` as a keyword, opposed to `weakSelf` which is colored as a local variable (i.e.: it's not colored).
Another reason is that I feel it's more explicit this way.
But if you guys insist I can change it.

Core/Core/Planner/CalendarToDo/ViewModel/CalendarToDoDetailsViewModel.swift
rh12: The assumption was that this screen is opened directly from a list which already contains the `plannable` and this update is there only for automatic updates after editing.
But you are right, it should be handled to make it more robust.
Updated to fallback to the input `plannable` in case of any error.

rh12:
@szabinst 
> I'm not sure how other errors (like non 200 http codes) are handled but without network connection there's no error handling.

My bad, I was mistaken about the error codes. Updated the PR description accordingly.

Will follow up on network connection loss and the offline mode topic.

rh12:
> Following other areas of the app, create and the edit button should not be enabled when offline mode is enabled.

Updated as per [discussion](https://instructure.slack.com/archives/G01GBRGPKQC/p1720185621928109?thread_ts=1720179500.670369&cid=G01GBRGPKQC):
- If internet goes off while a request is in-flight
  - show general error message (not a special message for no-internet)
- If OfflineModeFF is disabled AND internet is off
  - Create/Edit buttons stay enabled
    - on tap go to create/edit screen
  - Add/Save buttons stay enabled
    - on tap show general error message
- If OfflineModeFF is enabled AND internet is off
  - Create/Edit + Add/Save buttons are disabled (but tappable)
  - on tap show already standardized Offline Mode error message

Core/Core/InstUI/Views/Cells/TextFieldCell.swift
rh12: updated

vargaat (changes requested):
The add button's color seems to be off. This screenshot is from iOS 16.
IMG_0332

rh12:
> The add button's color seems to be off. This screenshot is from iOS 16.
Nice catch, thanks! Fixed it. Do you have any clue why did it occur only in iOS16?

vargaat (approved):
QA+1

vargaat:
> > The add button's color seems to be off. This screenshot is from iOS 16.
> 
> Nice catch, thanks! Fixed it. Do you have any clue why did it occur only in iOS16?

Most probably it was broken on iOS 17 as well, I tested it only on iOS 16 at that time.

szabinst (approved):
QA + 1

ndrsszsz (approved):
QA +1
Tested on iPhone 11, iOS 17.5.1

PR #2216, opened 2022-07-15
Can’t view documents embedded by students in a discussion

idevainstr:
refs: MBL-16152
affects: Student, Teacher
release note: Fixed can’t view documents embedded by students in a discussion.
test plan: See ticket.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

PR #2850, opened 2024-07-30
Fix 403 API responses triggering a token renewal

vargaat:
### What happened?
- Upon app start we initialize pandata that makes an API call to the backend to start a pandata session.
- In case an account has no pandata set up the API call is rejected with a 403 Developer key not registered.
- During a [bug fix](https://github.com/instructure/canvas-ios/pull/2216/files) in the past we mistakenly defined the 403 response code as unauthorised and thus eligible for a token refresh and retry.
- Upon logging into the no pandata account the app instantly renewed the access token.
- The old access token was invalidated since a new one was created and because of this in-flight API calls failed with error 401.
- These 401 API calls again renewed the access token causing another round of invalid access token API calls which renewed the token again.
- This cascade of token renewals quickly depleted the token refresh quota of the account so our requests automatically got rejected due to the frequent usage of this API.
- Our logic waits only 1 sec before retrying an API call that was rejected due to the quota being exhausted but this time wasn't enough for this account so the app tried to renew the token every 1 sec to infinity.

### Solutions
- There are multiple implementation quirks in the API.swift file that caused this event to happen. (For example API calls not being paused in case one was throttled down or the hardcoded 1 sec waiting time.) Fixing these would be much more complex and error prone than fixing the issue which triggered this event.
- I reverted the previous fix (linked above) and made the app not to renew the token in case a 403 was received. This broke the previous fix so I had to apply another -this time correct- fix.

refs: MBL-17768
release note: none

test plan:
- See ticket for test account.
- Log in a few times and check if there multiple calls to `/login/oauth2/token` after you have logged in.
- Since this fix modifies the logic of how file downloads are re-tried also test if a user's personal file added as an URL to a page opens correctly when logged in with a different user.
- For this log in with my student.
- Go to Pages and open "Mary's page".
- This contains a link to an image uploaded by another user.
- Tapping the link should open the image (instead of giving a permission error).

## Checklist

- [ ] Tested on phone
- [ ] Tested on tablet

### DIFF
--- a/Core/Core/API/APIError.swift
+++ b/Core/Core/API/APIError.swift
@@ -19,10 +19,10 @@
 import Foundation
 
 public enum HttpError {
-    static let unauthorized = 1000
-    static let forbidden = 1001
-    static let notFound = 1002
-    static let unexpected = 2000
+    public static let unauthorized = 1000
+    public static let forbidden = 1001
+    public static let notFound = 1002
+    public static let unexpected = 2000
 }
 
 public enum APIError: LocalizedError {
--- a/Core/Core/Extensions/ErrorExtensions.swift
+++ b/Core/Core/Extensions/ErrorExtensions.swift
@@ -19,9 +19,16 @@
 import Foundation
 
 public extension Error {
-
+    
     var isFrameLoadInterrupted: Bool {
-        let nsError = self as NSError
-        return nsError.domain == "WebKitErrorDomain" && nsError.code == 102
+        nsError.domain == "WebKitErrorDomain" && nsError.code == 102
+    }
+
+    var isForbidden: Bool {
+        nsError.domain == NSError.Constants.domain && nsError.code == HttpError.forbidden
+    }
+
+    private var nsError: NSError {
+        self as NSError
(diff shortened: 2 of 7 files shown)


inst-danger:

</details>

inst-danger:

</details>

inst-danger:

</details>

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>This pull request will not generate a release note.</td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Student, Teacher, Parent

[MBL-17768](https://instructure.atlassian.net/browse/MBL-17768)

**Canvas iOS** | 91.63% | 91.63% | 0.01%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ff3bde5c0b304661190de68a11b25c071f1751c7
</p>

szabinst (approved):
QA + 1

PR #2883, opened 2024-09-06
Update video poster generation logic

vargaat:
### What happened?
- Studio videos failed to sync in case a video had only audio track.
- The reason was that the thumbnail generation failed for such videos that interrupted the studio sync as a whole.

### What changed?
- I updated the thumbnail generation logic so it can't make the outer reactive stream fail.
- Errors instead are reported to Crashlytics for this step.
- The thumbnail generation error for audio only videos is swallowed now, it's safe to ignore it.
- I unified the error descriptions so they will be properly grouped in Crashlytics.
- The thumbnail generation code was extracted to a standalone entity.

refs: MBL-17853
affects: Student
release note: Fixed Studio videos not loading in offline mode in case a video only has audio track.

test plan:
- See ticket for account details.
- Sync the course containing audio only studio video.
- Go offline mode.
- Studio videos should load.

## Checklist

- [ ] Tested on phone
- [ ] Tested on tablet

### DIFF
--- a/Core/Core/Studio/Model/StudioAPIAuthInteractor.swift
+++ b/Core/Core/Studio/Model/StudioAPIAuthInteractor.swift
@@ -20,12 +20,16 @@ import Combine
 import CombineExt
 import WebKit
 
-public enum StudioAPIAuthError: Error {
+public enum StudioAPIAuthError: String, Error {
     case failedToGetLTIs
     case failedToGetTokenFromWebView
     case failedToGetAPIBaseURL
     case studioLTINotFound
     case failedToGetLaunchURL
+
+    var localizedDescription: String {
+        "\(Self.self).\(self)"
+    }
 }
 
 public protocol StudioAPIAuthInteractor {
--- a/Core/Core/Studio/Model/StudioVideoPosterInteractor.swift
+++ b/Core/Core/Studio/Model/StudioVideoPosterInteractor.swift
@@ -1,0 +1,81 @@
+//
+// This file is part of Canvas.
+// Copyright (C) 2024-present  Instructure, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Foundation
+
+public protocol StudioVideoPosterInteractor {
+
+    /// - returns: The URL of the poster file. Nil if it wasn't created due to an error.
+    func createVideoPosterIfNeeded(
+        isVideoCached: Bool,
+        mediaFolder: URL,
+        videoFile: URL
+    ) -> URL?
+}
+
+public class StudioVideoPosterInteractorLive: StudioVideoPosterInteractor {
+    public typealias PosterFactory = (_ videoFile: URL, _ posterLocation: URL) throws -> Void
+
+    private let analytics: Analytics
+    private let posterFactory: PosterFactory
+
+    /// - parameters:
+    ///   - analytics: Since poster creation is not a blocker step we don't throw any errors from here only report them to analytics.
+    ///   - posterFactory: For testing purposes. Live implementation should have a default value for this parameter.
+    public init(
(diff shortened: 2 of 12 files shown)


inst-danger:

</details>

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Fixed Studio videos not loading in offline mode in case a video only has audio track.

#### Affected Apps: Student

[MBL-17853](https://instructure.atlassian.net/browse/MBL-17853)

<details><summary>:x: XCTest failed: CoreTests/StudioVideoPosterInteractorLiveTests/testSkipsPosterGenerationIfVideoIsCached</summary>


    XCTAssertEqual failed: ("Optional(file:///Users/vagrant/Library/Developer/CoreSimulator/Devices/46C95008-E0C7-49E6-A4C1-8EFF7538A1C1/data/Containers/Data/Application/3BD672A6-28C6-4EEF-B154-B48595FC7E1E/tmp/CoreTestCase/123/poster.png)") is not equal to ("nil")
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8b09863f032dd16870419c2b2fce2336bd225028
</p>

szabinst (commented):
I synced with an older app version with Tamas's account linked in the ticket and it successfully completed. (It should have failed if I'm correct.)

When I installed this version and tried to sync, the sync starts but progress doesn't show on the dashboard and I just get a notification when it's done. 

Tried logging out, reinstalling, didn't help.

vargaat:
> I synced with an older app version with Tamas's account linked in the ticket and it successfully completed. (It should have failed if I'm correct.)

The studio sync failure is not reflected in the overall offline sync status. In this case the offline sync successfully completes while the studio sync fails in the background. If you check the studio video in offline mode it shouldn't display with an older app version. An error log is also generated in the debug menu for the failed studio sync.

> When I installed this version and tried to sync, the sync starts but progress doesn't show on the dashboard and I just get a notification when it's done.
> 
> Tried logging out, reinstalling, didn't help.

I've also seen such behavior earlier. For me the sync banner didn't display and one thread was waiting on a managed object context to perform a write block on the context's queue. We could investigate this further if it's a common issue.

szabinst:
> The studio sync failure is not reflected in the overall offline sync status.

Right, I forgot that part. 

About the banner not showing up, this is my first time seeing this behaviour so I thought this PR might have something to do with it.

Core/Core/Studio/Model/StudioAPIAuthInteractor.swift
```diff
@@ -20,12 +20,16 @@ import Combine
...
+public enum StudioAPIAuthError: String, Error {
     case failedToGetLTIs
     case failedToGetTokenFromWebView
     case failedToGetAPIBaseURL
     case studioLTINotFound
     case failedToGetLaunchURL
+
+    var localizedDescription: String {
+        "\(Self.self).\(self)"
+    }
```
szabinst, line 32: I would rather call this `stringValue` or something because localizedDescription is usually a full sentence that could be even used for user facing error handling.

szabinst (changes requested):
Functionality wise works good, I only had 1 remark.

Core/Core/Studio/Model/StudioVideoPosterInteractor.swift
```diff
@@ -0,0 +1,81 @@
...
+        if isVideoCached {
+            return posterLocation
+        }
+
+        do {
+            try posterFactory(videoFile, posterLocation)
+        } catch let error {
+            if error.isSourceTrackMissing == false {
+                // report
+                analytics.logError(name: "asd")
```
rh12, line 67: `"asd"`?

vargaat: Ooops.

Core/Core/Studio/Model/StudioAPIAuthInteractor.swift
vargaat: Good thinking. The main reason I chose this is because every Error has the `localizedDescription` field so when you catch unknown error types you are still safe to use this field. If we create a custom property you have to know if this error type will be thrown or not at a higher level and check for its type so you can call the error's custom property. In this case this error is just one of the many types that can be caught in `CourseSyncStudioMediaInteractorLive`. I think we are on the safe side regarding not putting this error in front of users since we rarely put errors directly in front of them mainly because `localizedDescription`s are awfully formatted and not user friendly.
What I could come up with is an error protocol that has an `analyticsDescription` or something similar so handling of all the different types would be reduced to one but we would still be required to add this custom logic to every place where we want to extract the error's description.

szabinst (approved):
QA + 1

rh12 (approved):
Code review +1

PR #2886, opened 2024-09-10
Demo Component: Selection Drop Down

suhaibabsi-inst:
This PR is to demo the amount of code required to consolidate selection components of Custom Picker screen into a single one. 

#### One important note:
For single selection component, if we are aiming to reduce the code footprint in the calling sites by using `DropDownSelector` variation, we might end up sacrificing animation as we are relying on using `fullScreenCover` to bring up a global screen-wide container. This is open for other proposals if it serves the goal. 

#### Before

https://github.com/user-attachments/assets/2b66b3a7-d7ef-42aa-8376-f88177539832

#### After

https://github.com/user-attachments/assets/f1721c80-f93d-4ab9-8a71-863b4c0a48f6


suhaibabsi-inst:
Please note that this note a proposal. This is demo to weigh in positives and negatives of using such approach.
Personally, I would always prefer using native components as much as possible, and even when it means repeating some of the code to adapt to a custom context (`SelectionMenu` as an example.). 

But would go on developing custom component at absolute necessity.

suhaibabsi-inst:
Closing as the original PR was merged.

PR #2890, opened 2024-09-13
Calendar Event Edit confirmation

rh12:
refs: [MBL-17861](https://instructure.atlassian.net/browse/MBL-17861)
affects: Student
release note: none

## What's changed
Added confirmation for updating recurring calendar events.
Some cases are considered invalid, and backend responds via HTTP status code 400 and a message in such cases.
These localized, user facing error messages are displayed to the user instead of the generic save error message.
Some of the cases (probably not a full list, that's why I didn't replicate this logic on client side):
- Change recurrence rule on just one instance of an event
- Update start date on all instances of an event

## Test plan
- Verify edit confirmation works just like on the web
  - different options for first event and the rest
  - (localized) error messages (see cases above and screenshots below)
- Verify confirmation is only shown for recurring events and only when editing them, not adding
- Verify changed events update in the calendar properly

## Screenshots
<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist
- [ ] A11y checked
- [x] Tested on phone
- [ ] Tested on tablet
- [x] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-17861]: https://instructure.atlassian.net/browse/MBL-17861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Planner/CalendarEvent/Model/API/APICalendarEventRequestBody.swift
+++ b/Core/Core/Planner/CalendarEvent/Model/API/APICalendarEventRequestBody.swift
@@ -55,6 +55,7 @@ struct APICalendarEventRequestBody: Codable, Equatable {
     }
 
     let calendar_event: CalendarEvent
+    let which: APICalendarEventSeriesModificationType?
 }
 
 #if DEBUG
@@ -68,7 +69,8 @@ extension APICalendarEventRequestBody {
         location_name: String? = nil,
         location_address: String? = nil,
         time_zone_edited: String? = nil,
-        rrule: RecurrenceRule? = nil
+        rrule: RecurrenceRule? = nil,
+        which: APICalendarEventSeriesModificationType? = nil
     ) -> PostCalendarEventRequest.Body {
         .init(
             calendar_event: .init(
@@ -81,7 +83,8 @@ extension APICalendarEventRequestBody {
                 location_address: location_address,
                 time_zone_edited: time_zone_edited,
                 rrule: rrule
-            )
+            ),
+            which: which
         )
     }
 }
--- a/Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
+++ b/Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
@@ -74,17 +74,17 @@ public final class CalendarEventDetailsViewModel: ObservableObject {
         self.completion = completion
         self.deleteConfirmation = deleteSingleItemConfirmation
 
-        onEventSetUpdateDeleteConfirmationModel()
-        onEventSetUpdateMenuButtonVisibility(userId: userId)
+        updateDeleteConfirmationModel(onSet: event)
+        updateMenuButtonVisibility(onSet: event, userId: userId)
 
         loadData()
 
-        onEditButtonTapShowEditScreen()
-        onDeleteButtonTapDeleteEventAfterConfirmation()
+        showEditScreen(on: didTapEdit)
+        deleteEventAfterConfirmation(on: didTapDelete)
     }
 
-    private func onEventSetUpdateDeleteConfirmationModel() {
-        event
+    private func updateDeleteConfirmationModel(onSet subject: CurrentValueSubject<CalendarEvent?, Never>) {
+        subject
             .compactMap { $0 }
             .sink { [weak self] in
                 guard let self else { return }
@@ -93,8 +93,8 @@ public final class CalendarEventDetailsViewModel: ObservableObject {
             .store(in: &subscriptions)
     }
 
-    private func onEventSetUpdateMenuButtonVisibility(userId: String) {
-        event
+    private func updateMenuButtonVisibility(onSet subject: CurrentValueSubject<CalendarEvent?, Never>, userId: String) {
+        subject
             .compactMap { $0 }
             .flatMap { [weak self] in
                 return self?.canManageEvent(context: $0.context, userId: userId)
@@ -106,14 +106,14 @@ public final class CalendarEventDetailsViewModel: ObservableObject {
             .store(in: &subscriptions)
     }
 
-    private func onEditButtonTapShowEditScreen() {
-        didTapEdit
+    private func showEditScreen(on subject: PassthroughSubject<WeakViewController, Never>) {
+        subject
             .sink { [weak self] in self?.showEditScreen(from: $0) }
             .store(in: &subscriptions)
     }
 
-    private func onDeleteButtonTapDeleteEventAfterConfirmation() {
-        didTapDelete
+    private func deleteEventAfterConfirmation(on subject: PassthroughSubject<WeakViewController, Never>) {
+        subject
             .map { [weak self] in
                 self?.shouldShowDeleteConfirmation = true
                 return $0
@@ -122,7 +122,7 @@ public final class CalendarEventDetailsViewModel: ObservableObject {
                 self?.deleteConfirmation.userConfirmsOption(passdownValue: $0)
                     ?? Empty().eraseToAnyPublisher()
             }
-            .sink { [weak self] in self?.deleteToDo(option: $0.0, from: $0.1) }
+            .sink { [weak self] (option, source) in self?.deleteEvent(option: option, from: source) }
             .store(in: &subscriptions)
     }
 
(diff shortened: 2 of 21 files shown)


inst-danger:

</details>

Core/Core/Planner/CalendarEvent/ViewModel/EditCalendarEventViewModel.swift
```diff
@@ -357,14 +353,77 @@ final class EditCalendarEventViewModel: ObservableObject {
...
+                return editConfirmation.userConfirmsOption()
+                    .map { $0 }
+                    .eraseToAnyPublisher()
+            }
+            .map { [weak self] in
+                self?.state = .data(loadingOverlay: true)
+                return $0
+            }
+            .flatMap { [weak self] in
+                (self?.saveAction(seriesModificationType: $0) ?? Empty().eraseToAnyPublisher())
```
suhaibabsi-inst, line 404: How about making `saveAction()` to return non-optional publisher, with `Empty()` if model doesn't exists.

```swift
private func saveAction(seriesModificationType: SeriesModificationType?) -> AnyPublisher<Void, Error> {
    guard let model else { return Empty().eraseToAnyPublisher() }
    ...
```

Core/Core/Planner/CalendarEvent/ViewModel/EditCalendarEventViewModel.swift
```diff
@@ -82,14 +84,19 @@ final class EditCalendarEventViewModel: ObservableObject {
...
                 case .add: String(localized: "We couldn't add your Event at this time. You can try it again.", bundle: .core)
                 case .edit: String(localized: "We couldn't save your Event at this time. You can try it again.", bundle: .core)
                 }
             }(),
             buttonTitle: String(localized: "OK", bundle: .core)
         )
-    }()
+    }
+
+    var editConfirmation: ConfirmationViewModel<SeriesModificationType> = .init()
```
suhaibabsi-inst, line 99:
```suggestion
    var editConfirmation = ConfirmationViewModel<SeriesModificationType>()
```

Core/Core/Planner/CalendarEvent/ViewModel/EditCalendarEventViewModel.swift
```diff
@@ -108,6 +115,8 @@ final class EditCalendarEventViewModel: ObservableObject {
     private var selectedCalendar = CurrentValueSubject<CDCalendarFilterEntry?, Never>(nil)
     /// Returns true if any of the fields had been modified once by the user. It doesn't compare values.
     private var isFieldsTouched: Bool = false
+    private let wasEventPartOfSeries: Bool
```
suhaibabsi-inst, line 118: Why not `isEventPartOfSeries`?

Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
```diff
@@ -93,8 +93,8 @@ public final class CalendarEventDetailsViewModel: ObservableObject {
             .store(in: &subscriptions)
     }
 
-    private func onEventSetUpdateMenuButtonVisibility(userId: String) {
-        event
+    private func updateMenuButtonVisibility(onSet subject: CurrentValueSubject<CalendarEvent?, Never>, userId: String) {
```
suhaibabsi-inst, line 96: the argument `onSet` is quite confusing to me. It makes expect that some `Set` value is being passed to it. 
Can we rename to `on` as we have on other methods.

Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
```diff
@@ -74,17 +74,17 @@ public final class CalendarEventDetailsViewModel: ObservableObject {
...
 
-        onEditButtonTapShowEditScreen()
-        onDeleteButtonTapDeleteEventAfterConfirmation()
+        showEditScreen(on: didTapEdit)
+        deleteEventAfterConfirmation(on: didTapDelete)
     }
 
-    private func onEventSetUpdateDeleteConfirmationModel() {
-        event
+    private func updateDeleteConfirmationModel(onSet subject: CurrentValueSubject<CalendarEvent?, Never>) {
```
suhaibabsi-inst, line 86: Here as well, can we rename argument name to be `on` rather `onSet`. 
Or, `onSubject` for more expressive one.

suhaibabsi-inst (commented):
Thanks you @rh12 !
I left few minor comments for your consideration.

Core/Core/Planner/CalendarEvent/Model/API/APICalendarEventRequestBody.swift
```diff
@@ -55,6 +55,7 @@ struct APICalendarEventRequestBody: Codable, Equatable {
     }
 
     let calendar_event: CalendarEvent
+    let which: APICalendarEventSeriesModificationType?
```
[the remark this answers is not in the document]
rh12: Since this is a DTO, here I would stick with the same naming backend uses

Core/Core/Planner/CalendarEvent/ViewModel/EditCalendarEventViewModel.swift
rh12: Because of the `weak self` we would still need to fallback to `Empty()` here as well. That's why I would keep it simply `nil` in the helper method's `guard`.

Core/Core/Planner/CalendarEvent/ViewModel/EditCalendarEventViewModel.swift
rh12: I tired to indicate (along with it being a `let`) that this is about the initial state of the event.
The present state may be different (ie.: user changes to "Does not repeat")

Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
rh12: The distinction I was trying to make here, was that `on: ` is to be used for actions (like `didTapButton`), while `onSet:` or `onChange:` is to be used for what are basically values. Strictly speaking both kind are just publishers, but one conveys an action, the other an assignment or change of a value.
With `onSet` instead of `onChange`, I tried to be explicit that the update happens for each assignment, not just for changes. But I'm not fully against using just `onChange` in these value cases.
I decided against `onSetOf` and `onChangeOf`, but maybe that would make it more clear.

Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
rh12: Same as above

Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
suhaibabsi-inst: Yes, I think `onChange` would be more readable to me.

Core/Core/Planner/CalendarEvent/ViewModel/EditCalendarEventViewModel.swift
suhaibabsi-inst: Will this differentiation seems overkill to me. 😄
It is just the use of present tense is more common in code. If you were to reason about the use of this in a given context, it is very unlikely that you'd worry about how that was set in the past, rather what would be the value of it at present. In code, it could be a bad thing if a piece of it might have a wide implications outside it.

rh12: The point is, that we should not use the present value of `isPartOfSeries`.
Renaming to `isInitialEventPartOfSeries` for more clarity.

Core/Core/Planner/CalendarEvent/ViewModel/CalendarEventDetailsViewModel.swift
rh12: Updating to `onChangeOf` in both methods

suhaibabsi-inst (dismissed):
Thanks @rh12 !
Looks good to me.

vargaat (approved):
QA+1

ndrsszsz (approved):
QA +1

PR #2931, opened 2024-10-14
Enable add recipients when reply from web

Ahmed-Naguib93:
refs: [MBL-17961](https://instructure.atlassian.net/browse/MBL-17961)
affects: Teacher, Student
release note: Enabled adding recipients when replying from the web.

### Note
- I have checked the API documentation, but it's useless.
- Then I checked with the Android team, so we can do what they did.
- group_conversation must be true as static value.
- The 'Send individual message to each recipient' toggle will depend on bulk_message instead of group_conversation.

### Test plan:
1- Create a new message and enable the "Send individual message to each recipient" toggle.

2- Open the recipient's account on the web, try to reply, and add more recipients to it.

## Screenshots
Screenshot 2024-10-14 at 2 03 25 PM

## Checklist
- [ ] Tested on phone
- [ ] Tested in light mode

[MBL-17961]: https://instructure.atlassian.net/browse/MBL-17961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Core/Core/Conversations/APIConversation.swift
```diff
@@ -312,7 +312,7 @@ public struct PostConversationRequest: APIRequestable {
         public let media_comment_id: String?
         public let media_comment_type: MediaCommentType?
         public let attachment_ids: [String]?
-        public let group_conversation: Bool?
+        public let group_conversation: Bool = true
         public let force_new: Bool? = nil // Setting this seems to cause the api to ignore group_conversation
```
Ahmed-Naguib93, line 316: As per the discussion with the Android team, we must set group_conversation = true.

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Enabled add recipients when reply from web

#### Affected Apps: Teacher, Student

[MBL-17961](https://instructure.atlassian.net/browse/MBL-17961)

<details><summary>:x: XCTest failed: CoreTests/PlannerListViewControllerTests/testLayoutParentApp</summary>


    XCTAssertEqual failed: ("Optional("Event")") is not equal to ("Optional("Assignment")")
    XCTAssertEqual failed: ("Optional("Assignment")") is not equal to ("Optional("Event")")
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 079dd5618cd1229b6ee3ea7d577f87d7aa6fc5d7
</p>

Core/Core/Inbox/ComposeMessage/ViewModel/ComposeMessageViewModel.swift
```diff
@@ -430,16 +430,15 @@ final class ComposeMessageViewModel: ObservableObject {
             subject = title
         }
         let isExceedsRecipientsLimit = recipientIDs.count > maxRecipientCount
-        let groupConversation = isExceedsRecipientsLimit ? true : !sendIndividual
+        let groupConversation = isExceedsRecipientsLimit ? true : sendIndividual
```
rh12, line 433:
```suggestion
        let bulkMessage = isExceedsRecipientsLimit ? true : sendIndividual
```

rh12 (changes requested):
Sanity check: are we sure that with these changes, the toggle state during disabling and the info message still makes sense?

Core/Core/Conversations/APIConversation.swift
szabinst: Is there no use case where we might want this set to false?

Ahmed-Naguib93: @szabinst yes, we set it true for ever.

Ahmed-Naguib93:
> Sanity check: are we sure that with these changes, the toggle state during disabling and the info message still makes sense?

@rh12  Thanks for your comment. Yes, the logic has changed slightly after we set group_conversation = true permanently. Now, we only check for bulkMessage. I have tested sending messages to over 100 recipients as well as fewer, and it works fine.

Core/Core/Conversations/APIConversation.swift
rh12: Even then, shouldn't this `true` be set by the UseCase? And tested there?
cc @szabinst

@Ahmed-Naguib93 Could you please provide some explanation why is this always true and comment that in code? My understanding is the following:
`group_conversation` is set to `true` in all cases, for legacy reasons. This matches Android behavior.
`bulk_messages` governs whether a message is sent individually or not

szabinst: I'd also move this to the UseCase level and add comment.

rh12 (approved):
QA + 1

This PR also fixes the [issue](https://instructure.slack.com/archives/G01GBRGPKQC/p1729583780246509) where some new messages were put into existing, but unrelated threads.

vargaat (approved):
QA+1

PR #2986, opened 2024-11-26
Restrict disabled tabs

rh12:
refs: MBL-18057
affects: Student
release note: Fixed disabled tabs being available in some cases.

## What's changed
Routes which lead to disabled tabs are not opening anymore.
Behavior follows the website, including the enabling/disabling of some related pages.
The special cases are the following:
- when "Pages" tab is disabled we disable
  - `course/:courseID/wiki`
  - `course/:courseID/pages`
  - `course/:courseID/pages/front_page`
- there are some other cases but they involve internal URLs only

## Problems
- Our `router` doesn't know whether a route is for internal navigation only or not. This means we need to consider app-only URLs too when checking for disabled tabs. The known exceptions had been handled.
- Backend doesn't provide a list of disabled tabs. It just returns the currently enabled tabs. We also have no list of all the tabs and new ones could be added any time. Remote logging is in place for unexpected tab formats.
- We assume the following tab url format: `course/:courseId/tabname`, plus there are some known exceptions. When a non-tab URL is introduced in Canvas which matches this format, the app won't know about it and disable it. We can handle this once we are made aware.

## Test plan
- Verify disabled tabs are not available via
  - direct links on pages, messages, etc
  - push notifications
  - outside apps when app is open
  - outside apps when app is closed
- Verify all other kind of links are opening properly
- Verify links to another course's disabled tabs are disabled properly

## Checklist
- [ ] Tested on phone
- [ ] Tested on tablet

### DIFF
--- a/Core/Core/Router/CourseTabUrlInteractor.swift
+++ b/Core/Core/Router/CourseTabUrlInteractor.swift
@@ -17,23 +17,65 @@
 //
 
 import Foundation
+import Combine
 
 public final class CourseTabUrlInteractor {
 
+    private struct TabModel {
+        let id: String
+        let htmlUrl: String
+    }
+
     private var enabledTabsPerCourse: [Context: [String]] = [:]
+    private var subscriptions = Set<AnyCancellable>()
 
     public init() { }
 
-    public func setEnabledCourseTabs(_ tabs: [APITab], context: Context) {
-        guard context.contextType == .course else { return }
+    public func setupTabSubscription() {
+        let useCase = LocalUseCase<Tab>(scope: .all)
+        return ReactiveStore(useCase: useCase)
+            .getEntitiesFromDatabase(keepObservingDatabaseChanges: true)
+            .replaceError(with: [])
+            .sink { [weak self] tabs in
+                self?.tabsDidUpdate(tabs)
+            }
+            .store(in: &subscriptions)
+    }
 
-        var tabPaths = tabs.map { $0.html_url.absoluteString }
+    public func clearEnabledTabs() {
+        enabledTabsPerCourse = [:]
+    }
+
+    private func tabsDidUpdate(_ tabs: [Tab]) {
+        var tabModels: [Context: [TabModel]] = [:]
+        tabs.forEach { tab in
+            guard tab.context.contextType == .course, let htmlURL = tab.htmlURL else { return }
+
+            let context = tab.context
+            if !tabModels.keys.contains(context) {
+                tabModels[context] = []
+            }
+            tabModels[context]?.append(.init(id: tab.id, htmlUrl: htmlURL.absoluteString))
+        }
+
+        tabModels.forEach { context, tabs in
+            let tabPaths = pathsForTabs(tabs, context: context)
+            enabledTabsPerCourse[context] = tabPaths
+        }
+    }
+
+    private func pathsForTabs(_ tabs: [TabModel], context: Context) -> [String] {
+        // adding backend-provided paths from `html_url`
+        var tabPaths = tabs.map {
+            logPathFormatIfUnknown(for: $0)
+            return $0.htmlUrl
+        }
 
         // Adding extra paths for some tabs besides the ones from `html_url`.
         // Some have counterparts defined in `routes`, used only for in-app navigation or they are just legacy.
         // Some have related paths and they should be enabled/disabled together.
         tabs.forEach { tab in
-            switch tab.id.value {
+            switch tab.id {
             case "syllabus":
                 // proper path from `html_url`: "course/:courseID/assignments/syllabus"
                 // internal path from `routes`: "course/:courseID/syllabus"
@@ -55,11 +97,7 @@ public final class CourseTabUrlInteractor {
             }
         }
 
-        enabledTabsPerCourse[context] = tabPaths
-    }
-
-    public func clearEnabledCourseTabs() {
-        enabledTabsPerCourse = [:]
+        return tabPaths
     }
 
     /// Returns `true` if `url` is not a course tab URL OR it is but it's not in the list of enabled course tab URLs.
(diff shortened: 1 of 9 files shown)


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Fixed disabled tabs were available in some cases.

#### Affected Apps: Student

[MBL-18057](https://instructure.atlassian.net/browse/MBL-18057)

<details><summary>:x: Swift lint</summary>


    
    ----- xcbeautify -----
    Version: 2.11.0
    ----------------------
    
    ❌ /Users/vagrant/git/Core/CoreTests/Courses/CourseTests.swift:303:79: Trailing Comma Violation: Collection literals should not have trailing commas (trailing_comma)
    /Users/vagrant/git/Core/CoreTests/Router/CourseTabUrlInteractorTests.swift:235:19: error: Colon Spacing Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals (colon)
    
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e874e9b4a69cbb16cb0bdf904b4bb4f812ffc5e7
</p>

Core/Core/Router/CourseTabUrlInteractor.swift
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    private var enabledTabsPerCourse: [Context: [String]] = [:]
+    private var subscriptions = Set<AnyCancellable>()
+
+    public init() { }
+
+    public func setupTabSubscription() {
+        let useCase = LocalUseCase<Tab>(scope: .all)
+        return ReactiveStore(useCase: useCase)
```
vargaat, line 36: No need to return here.

Core/Core/Router/CourseTabUrlInteractor.swift
```diff
@@ -0,0 +1,173 @@
...
+            .getEntitiesFromDatabase(keepObservingDatabaseChanges: true)
+            .replaceError(with: [])
+            .sink { [weak self] tabs in
+                self?.tabsDidUpdate(tabs)
+            }
+            .store(in: &subscriptions)
```
vargaat, line 42: We are storing this subscription that never ends `keepObservingDatabaseChanges: true` so we should cancel this when the user logs out not to keep the old coredata context stack alive.

Core/Core/Router/CourseTabUrlInteractor.swift
```diff
@@ -0,0 +1,173 @@
...
+            default:
+                break
+            }
+        }
+
+        return tabPaths
+    }
+
+    /// Returns `true` if `url` is not a course tab URL OR it is but it's not in the list of enabled course tab URLs.
+    public func isAllowedUrl(_ url: URL) -> Bool {
```
vargaat, line 103: Let's move this public function next to other public functions.

Core/Core/Router/CourseTabUrlInteractor.swift
```diff
@@ -0,0 +1,173 @@
...
+            // example: "/courses/42/assignments/syllabus"
+            return parts.count == 4 && parts[3] == "syllabus"
+
+        case .frontPage:
+            // example: "/courses/42/pages/front_page"
+            return parts.count == 4 && parts[2] == "pages" && parts[3] == "front_page"
+
+        case .externalToolTab:
+            // example: "/courses/42/external_tools/1234"
+            return parts.count == 4 && parts[2] == "external_tools"
```
vargaat, line 170: Do we want to add a guard in case there's an invalid route format not to crash with the array indexing?

vargaat (changes requested):

Core/Core/Router/CourseTabUrlInteractor.swift
rh12: Unless I'm missing something, `parts.count == 4` before the `&&` ensures that we never reach the `parts[2]` check.

But on a related note, I will update the syllabus rule to include the `assignments` part, to avoid nonexistent corner case of `external_tools/syllabus` depending simply on ordering of the cases

vargaat: True that! I was looking for a guard like on the first case.

ndrsszsz (approved):
QA +1

vargaat (changes requested):
Seems like we have a false positive: I hid the assignments page and the course's home was set to a page that I cannot open because the app thinks it's hidden. Let's take a look on this.

rh12:
@vargaat 
> Seems like we have a false positive: I hid the assignments page and the course's home was set to a page that I cannot open because the app thinks it's hidden. Let's take a look on this.

Great find, thanks! This was caused by two issues:
1. relative URLs without starting slash were not handled correctly
2. Home screen should allow all URLs, even disabled tabs
  - this is how it works on the web
  - updated PR description with this
  - please check this during QA

suhaibabsi-inst (approved):
Tested the following scenarios:

- ✅  Verify all other kind of links are opening properly
- ✅  Verify links to another course's disabled tabs are disabled properly
- ✅  Verify tapping "Course Details > Home" opens the screen when it leads to a disabled tab

Verify disabled tabs are not available via:
- ✅  direct links on pages, messages, etc
- ✅  push notifications
- ✅   outside apps when app is open
- ✅   outside apps when app is closed
    
_Note:_
Deep links (tapping on a link from outside the app) is not working for me, and I noticed that the `.well-known/apple-app-site-association` file for my test account isn't configured to allow so.
I am not sure if we have access to change this or not. 
Tested those using `canvas-student` url scheme.

QA + 1

PR #2989, opened 2024-11-27
Fix issue of user not being able to submit assignment when logged in root account

suhaibabsi-inst:
refs: [MBL-18064](https://instructure.atlassian.net/browse/MBL-18064)
affects: Student,Teacher
release note: Fixes issue of user not being able to submit assignment when logged in root account

## What's changed

The main point of change of this PR resides in files:
[`AppEnvironmentOverride`](https://github.com/instructure/canvas-ios/pull/2989/files#diff-2151f2282f4d1aadb4e9ae981de9758a1f7355399f3f146e18e87e53a1b8d683)
[`CourseTabUrlInteractor`](https://github.com/instructure/canvas-ios/blob/9c831a1c7135a0724efaeb85b48161ad9d776906/Core/Core/Router/CourseTabUrlInteractor.swift#L117)
[`Router`](https://github.com/instructure/canvas-ios/blob/9c831a1c7135a0724efaeb85b48161ad9d776906/Core/Core/Router/Router.swift#L348)

This is to provide a clean way to pass a modified `AppEnvironment` with the correct baseURL that fits the current tab user is sitting at, which in this case is `Course > Assignments`

The rest are a consequence of those changes, which is basically to make sure views and classes are respecting what ever `AppEnvironment` instance is passed to them.

## Test Plan

See [ticket](https://instructure.atlassian.net/browse/MBL-18064) description.

Also, made testing for ordinary accounts for following use cases:
1. View Assignment details
2. View Submission details
3. New text and file assignment submission.
4. New text and file submission comment.
5. View submission comments, especially those with comments.

Also, testing for the following scenarios:
1. Tapping on external link.
2. Tapping on an internal link for an assignment, making sure submission work as well for both of cases.

## Screen Record

https://github.com/user-attachments/assets/4b0c8fe8-f8de-48a3-9789-e6eafdc8a047

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-18064]: https://instructure.atlassian.net/browse/MBL-18064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


inst-danger:

</details>

Core/Core/Assignments/AssignmentList/ViewModel/AssignmentCellViewModel.swift
```diff
@@ -21,10 +21,10 @@ import SwiftUI
 public class AssignmentCellViewModel: ObservableObject {
     public let assignment: Assignment
     public let courseColor: UIColor?
+    private let env: AppEnvironment
 
-    private let env = AppEnvironment.shared
-
-    public init(assignment: Assignment, courseColor: UIColor?) {
+    public init(env: AppEnvironment, assignment: Assignment, courseColor: UIColor?) {
```
rh12, line 26: I think we usually pass in the primary data/model first which is the main subject of the whole type.
Then the secondary ones.
And in the end we pass in the dependecies, like interactors, env, router, etc.
Maybe this is only me, but it would make sense to me to keep this ordering.
May look like nit, but I would rather call this out now, at the beginning of this approach.

The same stands for all the other places where we inject `env` in this PR. cc @vargaat for thoughts

vargaat: Yes, I agree, that bloated environment object is definitely not a primary dependency. I didn't notice this during my review but let's keep the priority order in the future.

vargaat (dismissed):
QA+1, I was able to submit to both consortia and my regular account various ways. What I noticed and I'm not sure if this can cause any issues or not is that when submitting via the share extension the file upload request still goes to the root account while the submission goes to the child domain. This is because the file upload logic for the share extension fetches the base url for its own but only for the last API request. On the image the non-selected rows are from the share extension and bottom ones from the app.
Screenshot 2024-12-04 at 16 04 54
I also managed to make the app crash two times during submitting files from the app. I couldn't reproduce it with Xcode and for some reason the crash symbolication didn't succeed so let's keep an eye out for this while testing the RC.

Core/Core/Assignments/AssignmentList/ViewModel/AssignmentCellViewModel.swift
suhaibabsi-inst: Yes, that's right.
If possible, I will have a follow up PR after release to address that.

vargaat (approved):
QA+1, file share now correctly uses the subdomain hosts for all API calls.

ndrsszsz (approved):
QA +1

PR #3034, opened 2024-12-17
[Horizon] Add navigation bar

Ahmed-Naguib93:
Navigation bar

https://instructure.atlassian.net/browse/CLX-282


inst-danger:

</details>

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



<details><summary>:x: XCTest failed: CoreTests/AssignmentCellViewModelTests/testSubmissionStatusAndIconAndColor</summary>


    XCTAssertEqual failed: ("#6a7883") is not equal to ("#697783")
    XCTAssertEqual failed: ("#03893d") is not equal to ("#03893c")
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0057e2451e1a1ec2141d28baf42f29c64e5eb7ac
</p>

Horizon/Horizon/Sources/Features/Dashboard/View/DashboardView.swift
```diff
@@ -50,53 +51,17 @@ struct DashboardView: View {
                 }
             }
         }
-        .navigationBarItems(leading: nameLabel)
-        .navigationBarItems(trailing: navBarIcons)
+        .navigationBarItems(leading: HorizonUI.NavigationBar.Leading(title: "Hi, Horizon App"))
+        .navigationBarItems(trailing: HorizonUI.NavigationBar.Trailing(onEvent: viewModel.onEvent))
         .scrollIndicators(.hidden, axes: .vertical)
         .background(Color.backgroundLight)
+        .onFirstAppear { viewModel.viewController =  viewController}
```
reabbotted, line 58: I don't love the pattern of requiring the view to set a property on the view model. I think it's better just to pass the viewController as an argument to the notebookDidTap  method. That way it's always clear what is required. Otherwise it can be unclear to someone using the view model that they need to set that property before it'll work.

Ahmed-Naguib93: Ok but i need to pass the closure direct to view model, because the navigation logic  must be set in viewModel.

packages/HorizonUI/Sources/HorizonUI/Sources/Components/NavigationBar/TrailingNavigationBar.swift
```diff
@@ -0,0 +1,90 @@
...
+
+        private let onEvent: (Event) -> Void
+
+        public init(onEvent: @escaping (Event) -> Void) {
+            self.onEvent = onEvent
+        }
+
+        public var body: some View {
+            HStack(spacing: .zero) {
+                Button {
```
reabbotted, line 37: These will be swapped out with the icon only buttons.

Horizon/Horizon/Sources/Features/Dashboard/View/DashboardView.swift
reabbotted: It would just be one of the parameters passed in the onEvent handler:

func onEvent(event: HorizonUI.NavigationBar.Trailing.Event, viewController: WeakViewController)

You might also consider creating a separate ViewModel to pair with the HorizonUI.NavigationBar.Trailing and putting the onEvent handler in there. Otherwise we must handle the navigation logic in the ViewModel for every View that uses the HorizonUI.NavigationBar.Trailing view.

reabbotted (dismissed):

Horizon/Horizon/Sources/Features/Dashboard/View/DashboardView.swift
Ahmed-Naguib93: No, HorizonUI is just for creating UI only and shouldn't have any logic in it.

packages/HorizonUI/Sources/HorizonUI/Sources/Components/NavigationBar/TrailingNavigationBar.swift
Ahmed-Naguib93: Yes, you are right, once the component finished will use it

Horizon/Horizon/Sources/Features/Dashboard/View/DashboardView.swift
szabinst: Since this component will be used only on the dashboard, I'd simply pass 3 named `() -> Void` closure to the `HorizonUI.NavigationBar.Trailing` initializer then we can have: 
```
viewModel.notebookDidTap(viewController)
viewModel.notificationDidTap(viewController)
...the rest
```

that is readable easily.

szabinst (changes requested):
Please see my comments below.

PR #3038, opened 2024-12-20
Filter screens consolidation

rh12:
refs: [MBL-18544](https://instructure.atlassian.net/browse/MBL-18544)
affects: Student, Teacher
release note: Improved various filter screens.

## What's changed
Added `SingleSelectionView` and `MultiSelectionView` components (and their viewmodels) to handle selections, typically found on Filter or Preferences screens.

Added `OptionItem` as the model for these, which is a stripped down, unified model for all option selections.
- It is intended to be created from and match with the given domain model.
- Helper types `SingleSelectionOptions` and `MultiSelectionOptions` group them together for simplified usage.
- Kept its `id` as a `String` instead of making it generic. For simplicity and because it covers most cases.
- It has no related protocol like `OptionItemConvertible` at the moment.

Updated the following screens to use these:
- Student Assignment List Preferences
- Teacher Assignment List Preferences
- Grade List Preferences
- Smart Search Preferences
- Calendar Selector
- Calendar Filter 

Consolidated UI in accordance with UX designer (Figma update is WIP)
- components (RadioButtonCell, CheckboxCell, ListSectionHeader, Divider)
- unified the above preferences screens (consistent wording, ordering, etc.)

## What's not changed
- List screens are not updated to use the updated components
- Colors are not following context color yet
- Navigation Bar (button fonts + colors, title fonts, dynamic text sizes)
- Grade List Preferences screen Done behaviour:
  - swipe-to-dismiss is not added yet
  - kept "Done" button disabling for now
- Did not refactor/rework types around the affected screens, tried to draw the line somewhere

## Test plan
Verify all the above screens work as before, including the interactions with their List screens.
There is no updated Figma for each of the screens, so please check the screenshots below for UI changes.

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist
- [x] A11y checked
- [x] Tested on phone
- [x] Tested on tablet
- [x] Tested in dark mode
- [x] Tested in light mode
- [x] Tested on iOS 18
- [x] Tested on iOS 17

[MBL-18544]: https://instructure.atlassian.net/browse/MBL-18544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Improved various filter screens.

[MBL-18544](https://instructure.atlassian.net/browse/MBL-18544)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ad28fb62a9b170c13e099b1fe39571b5a2b4d4b0
</p>

inst-danger:

</details>

inst-danger:

</details>

Core/Core/Common/CommonUI/OptionSelection/View/MultiSelectionView.swift
```diff
@@ -0,0 +1,147 @@
...
+                ForEach(viewModel.allOptions) { item in
+                    optionCell(with: item)
+                }
+            } header: {
+                if hasAllSelectionButton {
+                    InstUI.ListSectionHeader(
+                        title: title,
+                        buttonLabel: Text(viewModel.allSelectionButtonTitle),
+                        buttonAction: { viewModel.didTapAllSelectionButton.send() }
+                    )
```
rh12, line 76: Regarding VoiceOver support, I've explored various solutions aiming to make the section header title and the all selection button one accessibility element.
The main problem is that headers can be quickly navigated to using swipe up/down gestures, but button action (or custom actions) are simply not available when using this type of navigation. Actions are available when the header is navigated to using swipe left/right gestures. This would make navigation/actions very inconsistent and confusing. It is the same on iOS 17/18.
Another approach was to add both SelectAll and DeselectAll custom actions, but adding a custom action means there is a default action too, which can't be removed. ("Activate, Default"). In this case there is no meaningful default action for the header, and the options ["Select All", "Deselect All", "Activate, Default"] would be confusing as well.

I ended up keeping the header and the SelectAll/DeselectAll button separate, which allows quick consistent navigation via headers, and allows for a reliable button action which announces "All options selected/deselected" when activated.

suhaibabsi-inst (commented):
Code +1

Core/Core/Common/CommonUI/OptionSelection/View/SingleSelectionView.swift
```diff
@@ -0,0 +1,126 @@
...
+    @StateObject private var viewModel: SingleSelectionViewModel
+
+    private let title: String?
+    private let accessibilityIdentifier: String?
+
+    public init(
+        title: String?,
+        accessibilityIdentifier: String? = nil,
+        allOptions: [OptionItem],
+        selectedOption: CurrentValueSubject<OptionItem?, Never>
```
suhaibabsi-inst, line 34: So, to go along with common design decision for Apple native components like "Picker"
Can you provide an initializer variation with `Binding` for selected option?

suhaibabsi-inst: And I am wondering whether we should use Combine subjects to pass data through view hierarchy rather than rely on default SwiftUI data flow techniques (Binding & States)?

Core/Core/Common/CommonUI/OptionSelection/Model/OptionItem.swift
```diff
@@ -0,0 +1,68 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import SwiftUI
+
+/// Represents a selectable option and its properties required to display it.
+public struct OptionItem: Equatable, Hashable, Identifiable {
```
suhaibabsi-inst, line 22: Perhaps we can leave that to another occasion,  but I was thinking of using generics here to have associated value attached to the option. This can have good consequences on the design of selection API.

```swift
public struct OptionItem<Value: Equatable>: Equatable, ... {
     ...
     var value: Value
}
```

suhaibabsi-inst (approved):
Few comments for your consideration about the API design.
Those are not blocking, and handling them, if applicable, can be left to another occasion.

suhaibabsi-inst (approved):
QA +1

Tested **Student** & **Teacher** apps on **iOS 18.2.1**, iPhone XS.

Core/Core/Common/CommonUI/OptionSelection/View/SingleSelectionView.swift
rh12: Using `CurrentValueSubject` to pass data between viewmodels instead of `Binding` is the current approach we use in the project. I don't remember the exact cause behind this decision, but I seem to recall it was because bindings caused issues and/or complications used this way.
We can have a team discussion about it sometime, as the topic may be worth exploring.
For now, I would not introduce a (currently unused) extra initializer.

ndrsszsz (approved):
Very nice job with these improvements! Thanks! 👍

vargaat (approved):
QA+1

PR #3050, opened 2025-01-03
Fix tests

vargaat:
There were two failures:
- One related to old discussions that caused our custom JS code not to execute. It seems the issue was that we didn't wait until the directory load navigation finished and we started loading html content immediately. I created a helper loader method in CoreWebView that calls back with a completion block when the navigation finishes and used this in the discussion view controller to load the actual content only after the first load have finished. We use the synchronous file load method in other places in the app but I didn't touch those to reduce the scope. I could only reproduce this issue on iOS 18.2.
- The other issue was that SwiftUI colors created from the asset catalog color resource had slightly different hex codes compared to their UIColor counterpart. I don't know what caused this but I've modified the logic to create Colors from UIColors instead and that solved the issue, the root cause is unknown.

refs: MBL-18545
release note: none
test plan:
- Sanity test if colors are okay.
- Test if discussions/announcements still work in offline mode.

## Checklist

- [x] Tested on iOS 17
- [ ] Tested on iOS 18

### DIFF
--- a/Core/Core/CoreWebView/View/CoreWebView.swift
+++ b/Core/Core/CoreWebView/View/CoreWebView.swift
@@ -341,7 +341,6 @@ open class CoreWebView: WKWebView {
         fileLoadNavigation = loadFileURL(url, allowingReadAccessTo: urlDirectory)
     }
 
-
     private func checkFileLoadNavigationAndExecuteCallback(finishedNavigation: WKNavigation) {
         if finishedNavigation == fileLoadNavigation {
             fileLoadNavigation = nil
--- a/Core/Core/CoreWebView/View/CoreWebView.swift
+++ b/Core/Core/CoreWebView/View/CoreWebView.swift
@@ -324,10 +324,41 @@ open class CoreWebView: WKWebView {
     public func activateFullScreenSupport() {
         fullScreenVideoSupport = .init(webView: self)
     }
+
+    // MARK: - File URL Load
+
+    private var fileLoadNavigation: WKNavigation?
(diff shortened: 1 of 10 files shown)


inst-danger:

</details>

inst-danger:

</details>

inst-danger:

</details>

Core/Core/CoreWebView/View/CoreWebView.swift
```diff
@@ -399,13 +434,29 @@ extension CoreWebView: WKNavigationDelegate {
         features.forEach { $0.webView(webView, didFinish: navigation) }
     }
 
+    public func webView(
+        _ webView: WKWebView,
+        didFail navigation: WKNavigation!,
+        withError error: any Error
+    ) {
+        checkFileLoadNavigationAndExecuteCallback(finishedNavigation: navigation)
```
rh12, line 442: Is it okay to execute the callback on failures, too?
Even the parameter name implies a `finished` navigation.
Would it make sense to for the completion closure to accept a `Result<Void, Error>` or similar?
That way callers would need to explicitly ignore the result (like `{ _ in ... }`)

vargaat: It's okay to pass back failures to the caller because we are only interested if the navigation have finished with any result (so it's not in progress when you start another load). Actually, if we load a directory here just to grant permission to the webview for the files the load will always fail with a navigation cancelled error. I renamed the parameter of the private method and passed back the result since it doesn't hurt anyone and might come handy in the future.

rh12 (approved):
QA + 1

PR #3053, opened 2025-01-09
Use `short_name` instead of `name` for student-student interactions

Q-Bea:
Changes `user.name` to `user.short_name` (or `user.shortName`) on the PeopleList and PeopleDetails pages when a student is viewing the details of another student.

Rational:
- This mimics the behaviour of the canvas web client (at least in my experience)
- "Short name" can also be used a "preferred name" by individuals of various demographics. It is not necessarily safe to assume that ones given name (`.name`) corresponds with how they wish to be referred as (`.short_name`).

- Teachers will still see the name as before, but it may also be worth adding the ability for them to see both a student's preferred and given names. 

Note:
- This is my first time interacting with this codebase and there are probably other changes to be made in order to stay consistent. I don't know where to look, though, and I don't have the ability to test the app on my devices.

### DIFF
--- a/Core/Core/People/ContextCard/Course/ContextCardHeaderView.swift
+++ b/Core/Core/People/ContextCard/Course/ContextCardHeaderView.swift
@@ -27,14 +27,19 @@ struct ContextCardHeaderView: View {
 
     var body: some View {
         VStack(spacing: 10) {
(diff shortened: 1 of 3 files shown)


vargaat:
Thank you for this idea. While we can't merge this as-is, I'm happy to inform you that this particular change as being part of a larger initiative is already on our plate and will be delivered in the future for our apps.

PR #3061, opened 2025-01-14
Accessibility - Teacher - Calendar - Font size on Edit Event screen

ndrsszsz:
affects: Student, Teacher
release note: None
test plan: No text should get cut off on EditCalendarEventScreen when font size is increased to the maximum.

I'm not quite pleased with the solution so any suggestion is appreciated.

Before:

https://github.com/user-attachments/assets/36bebf75-d38e-4d7a-a92d-187b65ffc92c

After:

https://github.com/user-attachments/assets/791b8c7c-11e1-4157-bb27-d7a064f73276

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
+++ b/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
@@ -62,33 +62,17 @@ extension InstUI {
 
         public var body: some View {
             VStack(spacing: 0) {
-                HStack(spacing: InstUI.Styles.Padding.standard.rawValue) {
-                    if dynamicTypeSize > .accessibility3 {
-                        VStack(alignment: .leading) {
-                            label
-                                .textStyle(.cellLabel)
-
-                            if date != nil {
-                                datePicker
-                            } else {
-                                placeholderButtons
-                            }
-                        }
-                    } else {
-                        label
-                            .textStyle(.cellLabel)
-
-                        if date != nil {
-                            datePicker
-                        } else {
-                            placeholderButtons
-                        }
+                ViewThatFits {
+                    HStack(spacing: InstUI.Styles.Padding.standard.rawValue) {
+                        dateRow
                     }
-                    if isClearable {
-                        clearButton
+                    .frame(minHeight: 36) // To always have the same height despite datepicker visibility
+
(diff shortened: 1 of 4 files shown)


inst-danger:

</details>

inst-danger:

</details>

suhaibabsi-inst (approved):
QA +1

Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
```diff
@@ -63,15 +63,27 @@ extension InstUI {
         public var body: some View {
             VStack(spacing: 0) {
                 HStack(spacing: InstUI.Styles.Padding.standard.rawValue) {
-                    label
-                        .textStyle(.cellLabel)
-
-                    if date != nil {
-                        datePicker
+                    if dynamicTypeSize > .accessibility3 {
```
rh12, line 66: This limit seems arbitrary to me, but not sure whether is there a proper way to check when would we extend the screen size.
What happens when we have the clear button?

What do you think about another way, to add a `geometry: GeometryProxy` to `DatePickerCell` and add `.frame(maxWidth: geometry.size.width)` in there, in place of the `.lineLimit(0)`.
It's less pretty but maybe less hacky as well.
We could discuss it briefly with the others on Monday.

Please update the other screens which uses this cell as well (`EditCalendarToDoScreen`, `EditCustomFrequencyScreen`)

rh12 (changes requested):

Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
vargaat: We should definitely check `ViewThatFits` since that's the way Apple proposes for such use cases.

vargaat (commented):
There's a sweet spot where the date is cropped in a single line. If I increase the font size it will break into two lines but the date text will still start out of the screen. I guess the date text itself widens the stack out of screen and the label just aligns to it.
IMG_0063

PR #3100, opened 2025-01-30
[Horizon] Using the course progression graphql endpoint

reabbotted:
The course progression endpoint doesn't have all the data we need for the components, so we still need to call the rest api. Sad face :(.


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please add a release note. If no release note is wanted, use `none`. Example: `release note: Fixed a bug that prevented users from enjoying the app.`
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Please add which apps this change affects. Example: `affects: Teacher, Student` or `affects: none`
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Please add a reference to a Jira ticket. For example: `refs: MBL-10023`
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



<details><summary>:x: Swift lint</summary>


    
    ----- xcbeautify -----
    Version: 2.11.0
    ----------------------
    
    ❌ /Users/vagrant/git/Core/Core/Features/Courses/CourseProgression/GetCoursesProgressionRequest.swift:84:1: Trailing Whitespace Violation: Lines should not have trailing whitespace (trailing_whitespace)
    /Users/vagrant/git/Core/Core/Features/Courses/CourseProgression/CDCourseProgression.swift:22:1: error: Vertical Whitespace Violation: Limit vertical whitespace to a single empty line; currently 2 (vertical_whitespace)
    
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8d9251c58e2a1ee41a45f6a6962d95b3bb6c5edb
</p>

Horizon/Horizon/Sources/Features/Learn/View/CourseDetailsViewModel.swift
```diff
@@ -37,13 +38,22 @@ final class CourseDetailsViewModel: ObservableObject {
 
     init(
         router: Router,
-        course: HCourse,
-        onShowTabBar: @escaping (Bool) -> Void
+        courseId: String,
+        onShowTabBar: @escaping (Bool) -> Void,
+        getCoursesInteractor: GetCoursesInteractor = GetCoursesInteractorLive(),
+        course: HCourse? = nil
```
szabinst, line 44: Can you describe the situation where we have a `courseId` but we don't have `course`? I don't see it looking at the changes.

Horizon/Horizon/Sources/Features/Learn/View/CourseDetailsViewModel.swift
```diff
@@ -37,13 +38,22 @@ final class CourseDetailsViewModel: ObservableObject {
...
         self.router = router
-        self.course = course
         self.onShowTabBar = onShowTabBar
         self.state = .data
+        self.course = course ?? .init()
+
+        getCoursesInteractor.getCourse(id: courseId)
+            .sink { course in
+                guard let course = course else { return }
+                self.course = course
```
szabinst, line 54: We probably need to use weak self here.

szabinst (commented):
If I understood correctly this is wip, so I only left comments about the code.

szabinst:
1 more thing, include one commit with [ignore-commit-lint] message so danger doesn't throw errors.

Horizon/Horizon/Sources/Features/Learn/View/CourseDetailsViewModel.swift
reabbotted: The thought is if we have to deep link to the page where the courseID is provided through the URL but we haven't loaded the course. But I can remove it.

Core/Core/Features/Courses/CourseProgression/CDCourseProgression.swift
```diff
@@ -16,31 +16,117 @@
...
         return model
     }
+
+    private static func saveCourse(
+        enrollmentModelCourse: GetCoursesProgressionResponse.CourseModel,
+        id courseId: String,
+        in context: NSManagedObjectContext
+    ) -> Course {
+        let course: Course = context.first(where: #keyPath(Course.id), equals: courseId) ?? context.insert()
+        course.id = courseId
```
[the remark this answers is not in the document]
reabbotted: It wouldn't be different. The thought was just that if it had to do an insert. through context.insert, we'd need to set the courseID.

Horizon/Horizon/Sources/Common/Domain/GetCoursesInteractor.swift
```diff
@@ -22,76 +22,94 @@ import Core
...
-                    updatedCourse.progressState = HCourse.ProgressState(from: completionPercentage)
-                    return updatedCourse
-                }
-            }
+    func getCourse(id: String) -> AnyPublisher<HCourse?, Never> {
+        fetchCourses()
+            .map { $0.first { $0.id == id } }
+            .receive(on: scheduler)
             .eraseToAnyPublisher()
     }
```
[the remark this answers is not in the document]
reabbotted: is there any preferred order to be set on these results?

inst-danger:

</details>

Horizon/Horizon/Sources/Common/Domain/GetCoursesInteractor.swift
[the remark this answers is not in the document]
szabinst: For the beta there'll be only 1 course. Beyond that, we'd need to align with the team.

Core/Core/Features/Courses/CourseProgression/CDCourseProgression.swift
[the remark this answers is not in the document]
szabinst: Right, I didn't see the insert!

Horizon/Horizon/Sources/Features/Learn/View/CourseDetailsViewModel.swift
szabinst: Good point, I'd probably then just leave the course id and fetch it through the interactor. 
Also we need to move `self.state = .data` to the `sink {}` callback because it might take time to actually get the course and then we should preserve the loading state.

reabbotted: I originally did just the course ID, but when doing that, it makes it so there's a moment where the page is loading and it shows no data while you have to wait for it to load. It's not big. But I figure if we have the course data already loaded from a previous page, it's a nicer experience if it just shows immediately. It's a pattern I tend to use where I support both so that it can adapt to whatever data we happen to have available.

reabbotted: I should also mention there's a development benefit here as well. When developing and testing, you can configure your environment to start with the page you're developing instead of having to navigate there each time.

reabbotted:
I've added back in the ability to only specify the courseID and not the HCourse for the CourseDetails page. The reason why is the Dashboard and CourseList pages need less information about the module items than the CourseDetails page. The GetCoursesProgressionUseCase is using the GraphQL endpoint to fetch module and module item information, but the GraphQL endpoint doesn't return all of the data we need for the CourseDetails page. So if only 1 course is requested, the GetCoursesProgressionUseCase will additionally load all of the module item information needed. It doesn't make these calls when requesting the entire list of courses for performance.

ktheory:
> The course progression endpoint doesn't have all the data we need for the components, so we still need to call the rest api.

That's surprising. Is this the case in the web version too? (I don't see the web version making REST calls.) Is there a difference in how iOS fetches data vs the web version?

Horizon/Horizon/Sources/Features/Learn/View/CourseListViewModel.swift
```diff
@@ -25,35 +25,72 @@ final class CourseListViewModel: ObservableObject {
...
+                    id: $0.id,
+                    institutionName: $0.institutionName,
+                    name: $0.name,
+                    progress: $0.progress / 100.0,
+                    progressString: $0.progress.progressString,
+                    progressState: $0.progress.progressState
+                )
+            }
+        }
+    }
```
szabinst, line 47: As I see this property is unused so we could simply move the mapping to the `getCourses` stream, then assign the result to the courses property.

szabinst (dismissed):
I only had 1 suggestion, otherwise it works great!

Ahmed-Naguib93 (dismissed):

Horizon/Horizon/Sources/Features/Learn/View/CourseListViewModel.swift
reabbotted: good catch, i'll update that

szabinst (approved):

PR #3112, opened 2025-02-05
Fix student e2e tests

vargaat:
### What changed?
- Fixed / updated all failing Student E2E tests so all green now.
- CoreWebView didn't load string content in offline mode correctly, this made content not appearing in the test. I couldn't reproduce it on real device but fixed the issue and the test now passes.
- The terminal service we use to turn off internet didn't correctly recognize errors when a proxy was used and could report invalid network status because of that.
- I encapsulated the E2E login flow into an XCTest activity, so the account ID the test uses is easily readable and copiable  in the test log in case you want to log in to that account manually. 
- Discovered a rendering glitch on the grading period list filter screen. Since I couldn't reproduce but the test always ran into this I created a follow-up task [MBL-18572](https://instructure.atlassian.net/browse/MBL-18572) to investigate and fix it properly. For now I added a workaround to the test.
- Refactored `GradingPeriodsTests` into activities so the test flow became easier to follow in the test log thus making locate errors also easy. I suggest to follow this pattern in the future. <br/>
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

refs: [MBL-18564](https://instructure.atlassian.net/browse/MBL-18564)
affects: Student
release note: none

test plan: E2E tests should pass.

[MBL-18572]: https://instructure.atlassian.net/browse/MBL-18572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MBL-18564]: https://instructure.atlassian.net/browse/MBL-18564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


suhaibabsi-inst (approved):
@vargaat 
Looks good to me.
I like the use `XCTContext.runActivity(name:)` to group elements lookup. This is very helpful. Thank you!

ndrsszsz (approved):

PR #3128, opened 2025-02-10
Accessibility - Student - AssignmentDetails - Attempts date button traits handling

ndrsszsz:
### Accessibility - Student - AssignmentDetails - Attempts date button traits handling
refs: MBL-18457
affects: Student
release note: None
test plan: VoiceOver should read attempts date as a button only if it is really a button.

Previously that attemptsDateButton was read by VoiceOver as a "button" even if it was not active/clickable and that could be misleading to the user. Now if the button is not active then we set the accessibilityTraits of the button to staticText.

## Checklist

- [x] A11y checked
- [x] Tested on phone
- [ ] Tested on tablet


rh12 (approved):
QA + 1

PR #3153, opened 2025-02-13
Accessibility - Student - Focus voiceover on sync progress card

vargaat:
- The ticket's suggestion to announce "Sync successfully" after the user confirmed the sync start is misleading so instead I focus on the sync progress card that announces that the sync have been started and also about progress updates.

refs: [MBL-18375](https://instructure.atlassian.net/browse/MBL-18375)
affects: Student
release note: none

test plan:
- Start an offline sync.
- Voiceover focus should jump to the sync progress card after modals have been dismissed and shuld announce what's happening.

## Checklist

- [ ] Tested on phone
- [ ] Tested on tablet

### DIFF
--- a/Core/Core/Features/Dashboard/OfflineSync/View/DashboardOfflineSyncProgressCardView.swift
+++ b/Core/Core/Features/Dashboard/OfflineSync/View/DashboardOfflineSyncProgressCardView.swift
@@ -21,6 +21,7 @@ import SwiftUI
 struct DashboardOfflineSyncProgressCardView: View {
     @Environment(\.viewController) private var viewController
     @ObservedObject private var viewModel: DashboardOfflineSyncProgressCardViewModel
+    @AccessibilityFocusState private var isProgressCardFocused: Bool
 
     public init(viewModel: DashboardOfflineSyncProgressCardViewModel) {
         self.viewModel = viewModel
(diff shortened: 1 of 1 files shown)


inst-danger:

</details>

suhaibabsi-inst:
QA +1

Tested in iPhone XS, iOS 18.2.0, Student app.

suhaibabsi-inst (approved):

rh12 (approved):
QA + 1

rh12 (changes requested):
I've tried it again on iPad today and it doesn't work for some reason. Sync progress is not focused. It focuses on a Dahsboard Course Card instead.
And it's the same on iPhone as well, for some reason, even though it worked before...
Please doublecheck
iPad @ 17.2
iPhone @ 18.2

vargaat:
> I've tried it again on iPad today and it doesn't work for some reason. Sync progress is not focused. It focuses on a Dahsboard Course Card instead.
And it's the same on iPhone as well, for some reason, even though it worked before...

I added a 1 second delay, now it seems more stable.

rh12:
> > I've tried it again on iPad today and it doesn't work for some reason. Sync progress is not focused. It focuses on a Dahsboard Course Card instead.
> > And it's the same on iPhone as well, for some reason, even though it worked before...
> 
> I added a 1 second delay, now it seems more stable.

Now sometimes it reads the sync progress, sometimes only reads the course card.
Would it make sense to announce Sync started after the sync screen disappears, then try to focus on the progress, and announce success once that finishes? Just throwing ideas around

vargaat:
> Would it make sense to announce Sync started after the sync screen disappears, then try to focus on the progress, and announce success once that finishes? Just throwing ideas around

I kept the focus logic as-is but added a sync started announcement. The sync's end is communicated via a local notification.

rh12 (approved):
QA + 1

I don't get the "Sync started" announcement on iPhone @ 18.2, but the Sync progress gets focused now consistently, so I think that's good enough.
On iPad @ iOS 17.2 it's the other way around.

PR #3162, opened 2025-02-16
Accessibility - Student - Lists Count Announcement For Offline Content Screen

suhaibabsi-inst:
refs: MBL-18373
affects: Student
release note: None

## Test Plan

First: Navigate to "Offline Content" screen for all courses
1- Make sure to turn Mobile Offline mode for your test account.
2- On Student app, in Dashboard screen, tap on 3-dots button on the right-left corner. 
3- Choose "Manage Offline Content"

Second: Upon loading, make sure the first item of each expandable list has "_**List, ? items, ..**_" prefix read out when focused via VoiceOver.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/CourseSync/Common/ViewModel/OfflineListCellViewModel.swift
+++ b/Core/Core/Features/CourseSync/Common/ViewModel/OfflineListCellViewModel.swift
@@ -21,6 +21,7 @@ import SwiftUI
 class OfflineListCellViewModel: ObservableObject {
 
     let cellStyle: OfflineListCellView.ListCellStyle
+    let followingListCount: Int?
     let title: String
     let subtitle: String?
     let selectionState: OfflineListCellView.SelectionState
@@ -31,6 +32,7 @@ class OfflineListCellViewModel: ObservableObject {
     let state: OfflineListCellView.State
 
     init(cellStyle: OfflineListCellView.ListCellStyle,
+         followingListCount: Int? = nil,
          title: String,
          subtitle: String? = nil,
          selectionState: OfflineListCellView.SelectionState = .deselected,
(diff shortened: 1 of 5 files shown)


inst-danger:

</details>

Core/Core/Features/CourseSync/Common/ViewModel/OfflineListCellViewModel.swift
```diff
@@ -21,6 +21,7 @@ import SwiftUI
 class OfflineListCellViewModel: ObservableObject {
 
     let cellStyle: OfflineListCellView.ListCellStyle
+    let followingListCount: Int?
```
rh12, line 24: Could you please move this under `isCollapsed` and keep that order for the init as well?

This property name seems a bit off for me, maybe because this prefix text for the list is built in one of it's cells.
I think it may make sense to pass in just a `accessibilityLabelPrefix: String?`, and just join that with the rest (with a comma) if it's not nil.
The string could be built in `makeViewModelItems()` (where you currently pass in the `followingListCount`), using a helper method there.

Core/Core/Features/CourseSync/Common/ViewModel/OfflineListCellViewModel.swift
```diff
@@ -134,28 +137,31 @@ class OfflineListCellViewModel: ObservableObject {
...
         }
 
         if !progressText.isEmpty {
-            finalText.append(", ")
-            finalText.append(progressText)
+            finalTextParts.append(progressText)
         }
 
-        return finalText
+        return finalTextParts.joined(separator: ", ")
```
rh12, line 164: Just a thought for cases like this: since we have `.nilIfEmpty` we can simplify these a bit:
```
        let finalText = [
            titleText.nilIfEmpty,
            ...
            progressText.nilIfEmpty
        ].compactMap { $0 }.joined(separator: ", ")
```
But no need to change if you prefer this way, just a suggestion

rh12 (changes requested):

rh12 (approved):
QA + 1

Thanks for the changes!

PR #3189, opened 2025-02-26
Accessibility - Student/Teacher - Refocus VoiceOver to DatePicker after selection is finished

suhaibabsi-inst:
refs: [MBL-18312](https://instructure.atlassian.net/browse/MBL-18312)
affects: Student, Teacher
release note: None.

This solution can be considered as a "hacky" solution, as it is being dependent on a UIKit value (`popoverDismissRegionAccessibilityID`) that was not officially declared in documentation for that use. Though upon testing, it turned out it serves our need perfectly. 

The main challenge of this work is to detect when popover is being presented, and which UI element caused it. While there is no way (officially documented by Apple) to tell that for `DatePicker`, in order to solve this, I had to rely on tracking a fixed ID of contributing elements in a global list, and use that to distinguish the source of popover upon dismissal.

## Test Plan

When activating DatePicker fields using VoiceOver on `Edit Event` screen, focus must be moved to them upon finishing with selection.

## Checklist

- [x] A11y checked
- [x] Tested on phone
- [x] Tested on tablet
- [ ] Tested in dark mode
- [x] Tested in light mode
- [ ] Approve from product

[MBL-18312]: https://instructure.atlassian.net/browse/MBL-18312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
+++ b/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
@@ -38,7 +38,6 @@ extension InstUI {
         private let errorMessage: String?
         private let isClearable: Bool
 
-        @State private var trackingPopoverID = Foundation.UUID().uuidString
         @Binding private var date: Date?
 
         public init(
@@ -148,7 +147,7 @@ extension InstUI {
                 displayedComponents: components,
                 label: {}
             )
-            .accessibilityRefocusingOnPopoverDismissal(trackingPopoverID)
+            .accessibilityRefocusingOnPopoverDismissal()
         }
 
         @ViewBuilder
--- a/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
+++ b/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
@@ -37,8 +37,8 @@ extension InstUI {
         private let validUntil: Date
         private let errorMessage: String?
         private let isClearable: Bool
-        private let trackingPopoverID: String?
 
+        @State private var trackingPopoverID = Foundation.UUID().uuidString
         @Binding private var date: Date?
 
         public init(
@@ -49,7 +49,6 @@ extension InstUI {
             validFrom: Date = .distantPast,
             validUntil: Date = .distantFuture,
             errorMessage: String? = nil,
-            trackingPopoverID: String? = nil,
             isClearable: Bool = false
         ) {
             self.label = label
@@ -59,7 +58,6 @@ extension InstUI {
             self.validFrom = validFrom
             self.validUntil = validUntil
             self.errorMessage = errorMessage
-            self.trackingPopoverID = trackingPopoverID
             self.isClearable = isClearable
         }
 
@@ -144,18 +142,13 @@ extension InstUI {
             case .dateAndTime: [.date, .hourAndMinute]
             }
 
-            let datePickerView = DatePicker(
+            DatePicker(
                 selection: binding,
                 in: validFrom...validUntil,
                 displayedComponents: components,
                 label: {}
             )
-
-            if let trackingPopoverID {
-                datePickerView.accessibilityRefocusingOnPopoverDismissal(trackingPopoverID)
-            } else {
-                datePickerView
-            }
+            .accessibilityRefocusingOnPopoverDismissal(trackingPopoverID)
         }
 
         @ViewBuilder
--- a/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
+++ b/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
@@ -37,6 +37,7 @@ extension InstUI {
         private let validUntil: Date
         private let errorMessage: String?
         private let isClearable: Bool
+        private let trackingPopoverID: String?
 
         @Binding private var date: Date?
 
@@ -48,6 +49,7 @@ extension InstUI {
             validFrom: Date = .distantPast,
             validUntil: Date = .distantFuture,
             errorMessage: String? = nil,
+            trackingPopoverID: String? = nil,
             isClearable: Bool
         ) {
             self.label = label
@@ -57,6 +59,7 @@ extension InstUI {
             self.validFrom = validFrom
             self.validUntil = validUntil
             self.errorMessage = errorMessage
+            self.trackingPopoverID = trackingPopoverID
             self.isClearable = isClearable
         }
 
@@ -110,12 +113,19 @@ extension InstUI {
                 get: { date ?? defaultDate },
                 set: { newDate in date = newDate }
             )
-            DatePicker(
+
+            let datePickerView = DatePicker(
                 selection: binding,
                 in: validFrom...validUntil,
                 displayedComponents: components,
                 label: {}
             )
+
+            if let trackingPopoverID {
+                datePickerView.accessibilityRefocusingOnPopoverDismissal(trackingPopoverID)
+            } else {
+                datePickerView
+            }
         }
 
         private var components: DatePicker<Label>.Components {
--- a/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
+++ b/Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
@@ -144,7 +144,7 @@ extension InstUI {
             case .dateAndTime: [.date, .hourAndMinute]
             }
 
-            DatePicker(
+            let datePickerView = DatePicker(
                 selection: binding,
                 in: validFrom...validUntil,
                 displayedComponents: components,
--- a/Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
+++ b/Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
@@ -27,11 +27,9 @@ struct A11yRefocusingOnPopoverDismissalViewModifier: ViewModifier {
 
     @AccessibilityFocusState
     private var isAccFocused: Bool
-    private let trackingPopoverID: String
 
-    init(trackingPopoverID: String) {
-        self.trackingPopoverID = trackingPopoverID
-    }
+    @State
+    private var trackingPopoverID: String = Foundation.UUID().uuidString
 
     func body(content: Content) -> some View {
         content
@@ -69,11 +67,7 @@ struct A11yRefocusingOnPopoverDismissalViewModifier: ViewModifier {
 extension View {
 
     /// Refocuses VoiceOver after popover dismissal to the element which activated it. Common examples are `DatePicker`s & `Menu`s
-    /// - parameters:
-    ///    - trackingPopoverID: This ID is used to distinguish popover source when there are multiple elements on screen that can activate popovers. It is important to pass a _distinct_ & **stable** value even if you have one single element on screen that can activate popover.
-    public func accessibilityRefocusingOnPopoverDismissal(_ trackingPopoverID: String) -> some View {
-        modifier(
-            A11yRefocusingOnPopoverDismissalViewModifier(trackingPopoverID: trackingPopoverID)
-        )
+    public func accessibilityRefocusingOnPopoverDismissal() -> some View {
+        modifier(A11yRefocusingOnPopoverDismissalViewModifier())
     }
 }
--- a/Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
+++ b/Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
@@ -96,5 +96,5 @@ public enum A11YPopoverFocus {
     private class DefaultTracking: A11YPopoverFocusTracking {
         var lastFocused: String?
     }
-    public static var tracking: A11YPopoverFocusTracking = DefaultTracking()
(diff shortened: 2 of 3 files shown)


inst-danger:

</details>

inst-danger:

</details>

inst-danger:
#### Release Note: 
None.

[MBL-18312](https://instructure.atlassian.net/browse/MBL-18312)

**Canvas iOS** | 91.66% | 91.68% | -0.01%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8eb14c46517d139d792a18b2cb93c5d6629afa99
</p>

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
```diff
@@ -0,0 +1,82 @@
...
+
+import Foundation
+import Combine
+import SwiftUI
+
+struct A11yRefocusingOnPopoverDismissalViewModifier: ViewModifier {
+    // Apple use this value for area where user can tap to dismiss pop-overs
+    static let popoverDismissRegionAccessibilityID = "PopoverDismissRegion"
+
+    private static var latestFocused = [String]()
```
rh12, line 27: It seems to me only the last element is checked and only all elements are removed,
which means, unless I'm missing something, that this could be simply a `lastFocused: String?`

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    @AccessibilityFocusState
+    private var isAccFocused: Bool
+
+    private let trackingPopoverID: String
+    init(trackingPopoverID: String) {
```
rh12, line 33: Would it be possible to just use an UUID string instead of an ID created by us for each View?
If it works, create the UUID in this init instead of passing some string in.
It would also allow to simply call this modifier on the datePicker without the `if let trackingPopoverID` there.

If not for some reason, create the UUID in the DatePickerCell and pass that in.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
```diff
@@ -0,0 +1,82 @@
...
+                isAccFocused = true
+                Self.latestFocused.removeAll()
+            }
+    }
+
+    private var didDismissPopover: AnyPublisher<Void, Never> {
+        NotificationCenter
+            .default
+            .publisher(for: UIAccessibility.elementFocusedNotification)
+            .filter({ notif in
```
rh12, line 57:
```suggestion
            .filter { notification in
```

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
rh12, line 27: Are we sure that this being static could not cause issues?

I get why does it need to be stored somewhere "higher up", and to have a proper injected place for it would complicate things probably needlessly. Just wanted to doublecheck that we can't think of any potential problem because of it.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: @rh12

This purpose of this ID is to distinguish this specific element from others that share the same behavior, so for example in "Edit Event" screen, there are three date pickers in the same screen, all are expected to return VoiceOver focus to them after selection is dismissed.

For you suggestion to create UUIDs, actually it tried that before, and found out that we need for those IDs to be stable across UIView updates, and I noticed when I tried that before, that those element get re-created each time Date is updated. So, to avoid complicating that from that perspective, I'd give the option to developer to decide on those. But the most important point, ID stable must be guaranteed during the whole view lifecycle.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: To be honest, nothing major comes to my mind when I think about possible issues of that, given that this is specific to VoiceOver function.
I'd prefer to keep using a global version of what I call `focus tracking` until we come across a critical one.

suhaibabsi-inst: And given that this a hacky solution in the first place, that means we aren't using an official API for it to work, I'd prefer avoiding over-engineering things around it.

rh12: I agree to not over-engineer, "Just wanted to doublecheck that we can't think of any potential problem because of it."

Did you introduce `protocol A11YPopoverFocusTracking` because of this? If so and we don't use it (the default one is used in the `DatePickerCell`) then it seems to be the same than a simple `static var`.
If this is the case we could just return to the simple solution, I would also prefer that.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
rh12: Thanks for checking this! It's a shame.

I would still try to avoid having all kinds of names and IDs for a single DatePicker (or similar element), so another approach I would prefer is to use the `accessibilityIdentifier` instead of a separate `trackingPopoverID`, because we would need the a11yID anyway for E2E tests:
- add a `accessibilityIdentifier: String?` parameter for DatePickerCell (with no default value, so it can't be overlooked)
- use that in place of the current `trackingPopoverID`
- use the same string in an actual `accessibilityIdentifier` modifier for the cell itself (probably just below the `.accessibilityElement(children: .contain)`
- remove the current `.accessibilityIdentifier` modifier on the screen, replacing it with the parameter

On more minor complication is the case when we have Date and Time together, like on the EditToDo screen.
For this we could simply add some suffix like `-date` and `-time` to each trackingID (computed from the a11yID as described).

This could all be encapsulated in the `DatePickerCell` itself, and the screens should just set the a11yID via the parameter. For value you could replicate the ones used in EditToDo screen.

Please also update all `DatePickerCell` occurrences with this because this refocus "feature" is needed on all screens having it.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: I think I changed my mind about over-engineering 😄 
I got caught with the point about improving testability.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: > On more minor complication is the case when we have Date and Time together, like on the EditToDo screen.
> For this we could simply add some suffix like -date and -time to each trackingID (computed from the a11yID as described).

I think I don't understand this point, you mean like adding those suffix based on date picker mode intentionally,
by adding those what ever value passed to `accessibilityIdentifier` argument?

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: I think we can use design, in case we would like to have some local reasoning about tracking latest element that initiated a popup, I understand the concerns about using global values, and thought, at least, I could give the option to developer to decide on that.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
rh12: Sorry, I forgot that the two buttons of the date picker are handled together by the system.
In Date & Time cases the focus will return to the date button. That's good enough I think.
So please ignore this paragraph.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
```diff
@@ -0,0 +1,100 @@
...
+public protocol A11YPopoverFocusTracking: AnyObject {
+    var lastFocused: String? { get set }
+}
+
+public enum A11YPopoverFocus {
+    private class DefaultTracking: A11YPopoverFocusTracking {
+        var lastFocused: String?
+    }
+    public internal(set) static var tracking: A11YPopoverFocusTracking = DefaultTracking()
+}
```
vargaat, line 100: At this stage I find this abstraction unnecessary. If we can come up with a valid use-case for this that can appear in the future then we can leave it as-is but otherwise just use a static variable in the view modifier.

vargaat (changes requested):
Good solution for this problem! One thing we should keep in mind is that if I move the focus to a date cell (so we store it as last focused) and then move it to a button for example and that button opens a custom popover, then when we dismiss the popover the focus will move to the date cell instead of the button so in such cases we should use this modifier on the button as well.

suhaibabsi-inst:
> Good solution for this problem! One thing we should keep in mind is that if I move the focus to a date cell (so we store it as last focused) and then move it to a button for example and that button opens a custom popover, then when we dismiss the popover the focus will move to the date cell instead of the button so in such cases we should use this modifier on the button as well.

Although, `lastFocused` value get erased once the popover is dismissed.
But, I agree, all UI element that can show popup on a given screen must subscribe to this for it work with no issues. This includes: Menus & DatePickers

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: Actually, the only use case that comes to my mind for this is to provide a way for developer to control this for a given context. But like you said, we can have introduced later on need.

Core/Core/Common/CommonUI/InstUI/Views/Cells/DatePickerCell.swift
```diff
@@ -37,6 +37,7 @@ extension InstUI {
         private let validUntil: Date
         private let errorMessage: String?
         private let isClearable: Bool
+        private let trackingPopoverID: String?
```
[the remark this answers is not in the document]
suhaibabsi-inst: Yes, I just tried it. It works!.
I remember I tried that before but something was off to me 🤔 
Anyway, thanks for the tip. This changes everything! 😄

rh12:
> > Good solution for this problem! One thing we should keep in mind is that if I move the focus to a date cell (so we store it as last focused) and then move it to a button for example and that button opens a custom popover, then when we dismiss the popover the focus will move to the date cell instead of the button so in such cases we should use this modifier on the button as well.
> 
> Although, `lastFocused` value get erased once the popover is dismissed. But, I agree, all UI elements that can show popup on a given screen must subscribe to this for it work with no issues. This includes: Menus & DatePickers.

Would clearing `lastFocused` regardless of the guard solve this?
Based on the assumption that we can never have more than one popovers.
Didn't test this, just thinking out loud.
```
            .onReceive(didDismissPopover) {
                if Self.lastFocused == trackingPopoverID {
                    isAccFocused = true
                }
                Self.lastFocused = nil
            }
```

suhaibabsi-inst:
> > > Good solution for this problem! One thing we should keep in mind is that if I move the focus to a date cell (so we store it as last focused) and then move it to a button for example and that button opens a custom popover, then when we dismiss the popover the focus will move to the date cell instead of the button so in such cases we should use this modifier on the button as well.
> > 
> > 
> > Although, `lastFocused` value get erased once the popover is dismissed. But, I agree, all UI elements that can show popup on a given screen must subscribe to this for it work with no issues. This includes: Menus & DatePickers.
> 
> Would clearing `lastFocused` regardless of the guard solve this? Based on the assumption that we can never have more than one popovers. Didn't test this, just thinking out loud.
> 
> ```
>             .onReceive(didDismissPopover) {
>                 if Self.lastFocused == trackingPopoverID {
>                     isAccFocused = true
>                 }
>                 Self.lastFocused = nil
>             }
> ```

We can do that because this closure will be called on all elements subscribing to popover behavior. 
If you tried that, you'd notice an inconsistent behavior, which depends on the first element that gets that notification.

Core/Core/Common/CommonUI/ViewModifiers/A11yRefocusingOnPopoverDismissalViewModifier.swift
suhaibabsi-inst: @rh12 
If you don't mind, will keep the argument named `trackingPopoverID`.
After changes suggested by @vargaat , having a fixed UUID value with `@State`,  `DatePickerCell` no longer should worry about it.

rh12 (approved):
QA + 1

Thanks for finding this hidden solution!

vargaat (approved):
QA+1

PR #3193, opened 2025-02-27
Fix issue where Studio videos not playing properly on offline mode

suhaibabsi-inst:
refs: [MBL-18585](https://instructure.atlassian.net/browse/MBL-18585)
affects: Student, Teacher
release note: Fixes issue where downloaded Studio videos not being played properly on offline mode.

## Test Plan

- Start by creating a page on Web with a studio video embedded into it.
- Open Mobile app, then sync Offline content of the course which that page belongs to.
- Navigate in the app to that page on offline mode.
- You must be able to play the video on that page with no issues on multiple devices.

See [ticket's](https://instructure.atlassian.net/browse/MBL-18585) description for more information on that.

## Checklist

- [ ] A11y checked
- [x] Tested on phone
- [ ] Tested on tablet
- [x] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-18585]: https://instructure.atlassian.net/browse/MBL-18585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


inst-danger:

</details>

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Fixes issue where downloaded Studio videos not being played properly on offline mode.

[MBL-18585](https://instructure.atlassian.net/browse/MBL-18585)

<details><summary>:x: XCTest failed: CoreTests/StudioAPIAuthInteractorLiveTests/testAuthenticationFlow</summary>


    Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "Output received from publisher", "Publisher finished".
    XCTAssertEqual failed: ("false") is not equal to ("true")
    XCTAssertEqual failed: ("false") is not equal to ("true")
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against feea8582464e9de99251c890f2fb5942a0633fab
</p>

Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift
vargaat, line 1: Could you please elaborate on why this works? Webview permissions were kind of a gray area when I last checked. I'm also concerned about if this shared file will cause any issues or not since all webviews will read this and there could be cases when there are multiple webviews displaying offline content in parallel. (Like a page has a link to another page or any other screen with offline RCE content.)

suhaibabsi-inst: @vargaat 

[*] For your question about why this works, but not the current way?

Honestly, I couldn't find a definitive justification for that. (like a guide or an official document).
However, reading through documentation for both methods:

[`loadHTMLString(_:baseURL:)`](https://developer.apple.com/documentation/webkit/wkwebview/loadhtmlstring(_:baseurl:))
[`loadFileURL(_:allowingReadAccessTo:)`](https://developer.apple.com/documentation/webkit/wkwebview/loadfileurl(_:allowingreadaccessto:))

You can tell from the description for `allowingReadAccessTo` that this made specific for content served over file system,  while it was stated for `baseURL` that it's restricted to file or web content. It doesn't even say any thing about read access. 
And, given that `loadFileURL(..)` method was introduced in iOS 9+, and the other was in iOS 8+. I think this method was Apple answer for our use case.

[*] For the issue of serving multiple pages at the same time, thanks for pointing that out, I will work on a solution to manage multiple copies of these. And perhaps, we can try saving those generated files to cache directory rather than manage their lifecycle ourselves.

Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift
```diff
@@ -709,17 +722,41 @@ extension CoreWebView {
...
+            } catch {
+                RemoteLogger.shared.logError(
+                    name: "Loading Offline Page",
+                    reason: error.localizedDescription
+                )
+                return
+            }
+
             loadFileURL(
-                URL.Directories.documents,
```
vargaat, line 713: If I change this documents folder url to `filePath` on the `master` branch it seems to be working on my sandbox and also with the sandbox in the ticket (ons iOS 18.2), could you please check if this works for you as well? I also increased the 2 seconds timeout to 5.

suhaibabsi-inst: @vargaat 
Yes, it would definitely work.
But you miss the extra code added by `html(for content: String)` to the content of that file, which was automatically added when calling the overridden [`loadHTMLString(...)`](https://github.com/instructure/canvas-ios/blob/f64301aeeda594e3f951fb7c046289f28df1ad56/Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift#L177) method.

suhaibabsi-inst: That's make sense!

suhaibabsi-inst: Just check with both accounts, yours and the one posted on ticket, and they seem to work as expected.
I did cleanup for the PR to only change this line. Kept the check timeout as 5.
Thanks, @vargaat , for checking this out.

suhaibabsi-inst: @vargaat 
Using iOS 18.3.0
When I tested again for test account posted on ticket. The video won't playing!

suhaibabsi-inst: Images loads successfully though!

vargaat (approved):

ndrsszsz (approved):
QA +1

PR #3201, opened 2025-03-03
Ignore non tested entities

vargaat:
[ignore-commit-lint]


suhaibabsi-inst (approved):

Core/Core/Common/CommonUI/UIViews/HorizontalScrollingStackView.swift
```diff
@@ -18,7 +18,7 @@
 
 import UIKit
 
-public class HorizontalScrollingStackview: UIView {
+public class HorizontalScrollingStackView: UIView {
```
rh12, line 21: Based on this change, it's no longer used.
I think it was a leftover from the old Parent Inbox.
If this seems something we would use we could keep it, but probably we should simply remove it

rh12 (approved):

ndrsszsz (approved):

PR #3215, opened 2025-03-07
Accessibility - Add support for webviews to change their font size based on the device's setting

vargaat:
### What's new?
- There's a new `CoreWebView` feature to dynamically adjust the font size in of the html content based on the system's dynamic font size settings.
- Since dynamic size category is an enum without values (*) I had to come up with a mapping to manually figure out the font scale for each enum value. For this I created a preview and I tweaked the scales until it matched the system font's size.

refs: MBL-18628
affects: Student, Parent, Teacher
release note: none

test plan:
- Test if webviews follow the device's dynamic font size.
- Note that webviews doesn't dynamically refresh when the font size is changed while on screen.

(*) Actually we can see in the control center the associated percentage values for each category but if I resize the font by that amount they don't match the font size determined by the system when using the dynamic size enum. Maybe the font sizing in the preview isn't correct?

## Screenshots

<table>
<tr><th>Preview for mapping</th></tr>
<tr>
<td></td>
</tr>
</table>

## Checklist

- [ ] Tested on iOS 17
- [x] Tested on iOS 18

### DIFF
--- a/Core/Core/Common/CommonUI/CoreWebView/Model/DefaultWKWebViewConfiguration.swift
+++ b/Core/Core/Common/CommonUI/CoreWebView/Model/DefaultWKWebViewConfiguration.swift
@@ -36,5 +36,9 @@ public extension WKWebViewConfiguration {
         allowsPictureInPictureMediaPlayback = true
         allowsAirPlayForMediaPlayback = true
         processPool = CoreWebView.processPool
+
+        // This is to make -webkit-text-size-adjust work on iPads.
+        // https://trac.webkit.org/changeset/261940/webkit
+        defaultWebpagePreferences.preferredContentMode = .mobile
     }
 }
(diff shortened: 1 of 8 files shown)


inst-danger:

</details>

inst-danger:

</details>

inst-danger:

</details>

suhaibabsi-inst (approved):
QA +1

Tested on iPhone XS, iOS 18.3.0.

rh12 (dismissed):
QA + 1

I found one place where font size may grow too much, but maybe because it was created with in-app RCE.
On Calender Event Details screen compare Location/Address values with Details value:

<table>
<tr><th>100%</th><th>160%</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

rh12 (approved):
QA + 1, the issue above is fixed

PR #3245, opened 2025-03-17
Fix issue of inaccessible Media Files attachments of Discussions

suhaibabsi-inst:
refs: MBL-18667
affects: Student, Teacher
release note: Fixes issue of inaccessible Media Files attachments of Discussions.

## Approach

The issue was caused by the WebView blocking the streaming of video files if the response included the header `Content-Disposition: attachment; ...`.

So, the resolution was basically to download the attachment, then have that video loaded into web view from a local temporary URL. While downloading the attachment the spinner is kept on screen until is finished or failed. 

## Test Plan

See [ticket's](https://instructure.atlassian.net/browse/MBL-18667) description.
Note that for large video size, you'd expect the spinner to stay longer until you get to see video controls.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


inst-danger:

</details>

inst-danger:

</details>

Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift
```diff
@@ -729,3 +740,98 @@ extension CoreWebView {
...
+            headerValue(of: "content-type")
+        }
+
+        private func headerValue(of key: String) -> String? {
+            guard let headers = (response as? HTTPURLResponse)?.allHeaderFields else { return nil }
+            return headers
+                .first(where: { ($0.key as? String)?.lowercased() == key.lowercased() })?
+                .value as? String
+        }
+    }
```
vargaat, line 834: I suggest to move these directly to the `URLResponseExtensions` file without this `HTTPHeader` struct. We have such lookups in that extension and the `headerValue` method seems to be generic enough to be used elsewhere. Also we could use the `HTTPHeaderConstants` file if we have such.

Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift
```diff
@@ -729,3 +740,98 @@ extension CoreWebView {
         }
     }
 }
+
+// MARK: - Attachment Download
```
vargaat, line 744: Let's extract this logic to a separate file/files/folder. This CoreWebView file is already big enough.

vargaat (changes requested):
Nice fix! I've had some comments around code organization.

vargaat (approved):
QA+1

rh12 (approved):
QA + 1

PR #3274, opened 2025-03-27
Student - K5 courses shows hidden tabs

ndrsszsz:
## K5 courses shows hidden tabs

To get the correct tabs for a K5 course `course_subject_tabs` must be included with the request to the tabs endpoint.
When we tap on the course then the K5SubjectViewModel retrieves the tabs and it uses the correct request with this `course_subject_tabs` included. However, when we arrive to the course screen we already have the tabs in CoreData because we already sent a request to the courses endpoint at app start with tabs included but without this `course_subject_tabs`, so when the course opens then we will see all tabs and not only the ones for K5 because the correct request with the include is not being sent.

The fix is to include `course_subject_tabs` also in the courses request at app start.

_Note: With mixed accounts (for users enrolled in both k5 and normal courses) it will only display k5 tabs. Same is true for web app. The cause is that in these cases the API returns with only the k5 tabs no matter the course._

refs: MBL-18707
affects: Student
release note: Fixed a bug that caused hidden tabs being shown in elementary type courses.
test plan: Check if tabs are shown correctly. See ticket for details.

## Checklist

- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Courses/APICourse.swift
+++ b/Core/Core/Features/Courses/APICourse.swift
@@ -263,11 +263,11 @@ public struct GetCoursesRequest: APIRequestable {
         case sections
         case syllabus_body
         case tabs
+        case course_subject_tabs  // for k5 tabs
         case term
         case total_scores
         case settings
         case grading_scheme
-        case course_subject_tabs  // for k5 tabs
     }
 
     let enrollmentState: EnrollmentState?
@@ -299,8 +299,10 @@ public struct GetCoursesRequest: APIRequestable {
     }
 
     public var query: [APIQueryItem] {
-        let isK5 = AppEnvironment.shared.k5.isK5Enabled
-        let includes = isK5 ? Include.allCases : Include.allCases.filter { $0 != .course_subject_tabs }
+        var includes = Include.allCases
+        if AppEnvironment.shared.k5.isK5Enabled {
+            includes.removeAll { $0 == .course_subject_tabs }
+        }
         return [
             .include(includes.map { $0.rawValue }),
             .perPage(perPage),
--- a/Core/Core/Features/Courses/APICourse.swift
+++ b/Core/Core/Features/Courses/APICourse.swift
@@ -267,6 +267,7 @@ public struct GetCoursesRequest: APIRequestable {
         case total_scores
         case settings
         case grading_scheme
+        case course_subject_tabs  // for k5 tabs
     }
 
     let enrollmentState: EnrollmentState?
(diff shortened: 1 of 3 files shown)


inst-danger:

</details>

inst-danger:
#### Release Note: 
Fixed a bug that caused hidden tabs being shown in elementary type courses.

#### Affected Apps: Student

[MBL-18707](https://instructure.atlassian.net/browse/MBL-18707)

**Canvas iOS** | 91.62% | 91.63% | -0.01%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1d727b25401d257e6e450d19e9b57301a11c5e26
</p>

Core/Core/Features/Courses/APICourse.swift
```diff
@@ -267,6 +267,7 @@ public struct GetCoursesRequest: APIRequestable {
         case total_scores
         case settings
         case grading_scheme
+        case course_subject_tabs  // for k5 tabs
```
rh12, line 270: nit: I would move this just below `tabs`, to avoid overlooking any of them in case we check one or the other

rh12 (changes requested):
Overall it works, but it seems we display Smart Search tab which just shows the Home page. It is not visible on the web (even if it is not disabled/hidden in Settings > Navigation), I assume because it is not intended for K5.
based on this, we should remove it as well.
(BTW we also remove that tab for regular courses, although we do it because SmartSearch is in the navbar instead)

Core/Core/Features/Courses/APICourse.swift
```diff
@@ -298,8 +299,12 @@ public struct GetCoursesRequest: APIRequestable {
     }
 
     public var query: [APIQueryItem] {
-        [
-            .include(Include.allCases.map { $0.rawValue }),
+        var includes = Include.allCases
+        if AppEnvironment.shared.k5.isK5Enabled {
+            includes.removeAll { $0 == .course_subject_tabs }
+        }
```
vargaat, line 305: Hm, what will happen on accounts with mixed K5 and non K5 courses?

ndrsszsz: With mixed accounts (for users enrolled in both k5 and normal courses) it will only display k5 tabs. Same is true for web app. The cause is that in these cases the API returns with only the k5 tabs no matter the course. I also added this info to the PR description and the code as comment.

vargaat (approved):
QA+1 for hidden tabs not being displayed.

A groups tab is still visible (but that's visible on web as well). Not sure if we want to have it here or not since this was not in the scope of the original elementary view. If yes, we should assign it to a proper icon because the selection indicator is glitching on it.
Screenshot 2025-03-28 at 11 21 49

PR #3282, opened 2025-03-31
[Horizon] Notebook Design Review Updates

reabbotted:
1. The close button should have a drop shadow behind it

2. The icon and header look too small

3. The text fields and label buttons should have strokes, not drop shadows

4. There’s a “cancel” button beneath the Save that shouldn’t be there (and if it is, it shouldn’t be white)

5. It looks like I can currently select both the confusing and important labels

6. After saving a note, I should be returned to the learning object and see a success toast. It currently keeps me in the menu, removes the save button, and shows a delete button

I also rolled this ticket into the changes. The styling had changed quite a bit since the previous implementation so it's now been updated to match what's currently designed.

https://instructure.atlassian.net/browse/CLX-1086

https://github.com/user-attachments/assets/87b1e519-8108-4f47-856b-6da594d423dd

[ignore-commit-lint]

### DIFF
--- a/Horizon/Horizon/Resources/Localizable.xcstrings
+++ b/Horizon/Horizon/Resources/Localizable.xcstrings
@@ -948,6 +948,9 @@
     "Your draft was deleted." : {
 
     },
+    "Your note has been successfully saved" : {
+
+    },
     "Your Submission" : {
 
     },
--- a/Horizon/Horizon/Resources/Localizable.xcstrings
+++ b/Horizon/Horizon/Resources/Localizable.xcstrings
@@ -948,7 +948,7 @@
     "Your draft was deleted." : {
 
     },
-    "Your note has been successfully saved" : {
+    "Your note has been saved" : {
 
     },
     "Your Submission" : {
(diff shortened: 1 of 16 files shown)


inst-danger:

</details>

szabinst (commented):
Thanks for the fixes! 

I noticed a couple of things. Not sure if all of them were present previously. 

Selecting a filter sets a loading state that covers the whole screen. Because the filtering is done in a short amount of time it feels like the screen is flickering: 
https://github.com/user-attachments/assets/95fc2a24-0887-49e6-b054-d9dd85b467bf


Nav bar should have a solid background color: 
https://github.com/user-attachments/assets/5bc1ee93-9d39-4f64-bfdd-f136ef65476b

When you open notebook from a learning object, you can then open the note which would open the learning object (where the note was taken) with the whole module sequence screen. (and you could continue doing that) This also screws up the navigation bar appearance and the back button also stops working because it's a modal. 
https://github.com/user-attachments/assets/f49b2dee-ef58-480f-8129-3569727664f4

reabbotted:
> Thanks for the fixes!
> 
> I noticed a couple of things. Not sure if all of them were present previously.
> 
> Selecting a filter sets a loading state that covers the whole screen. Because the filtering is done in a short amount of time it feels like the screen is flickering: https://github.com/user-attachments/assets/95fc2a24-0887-49e6-b054-d9dd85b467bf
> 
> Nav bar should have a solid background color: https://github.com/user-attachments/assets/5bc1ee93-9d39-4f64-bfdd-f136ef65476b
> 
> When you open notebook from a learning object, you can then open the note which would open the learning object (where the note was taken) with the whole module sequence screen. (and you could continue doing that) This also screws up the navigation bar appearance and the back button also stops working because it's a modal. https://github.com/user-attachments/assets/f49b2dee-ef58-480f-8129-3569727664f4

Good call on all 3.

https://github.com/user-attachments/assets/3a276da3-56f3-427c-ad37-732caa693cf5

szabinst (approved):

PR #3289, opened 2025-04-04
[Horizon] Display Course Score

reabbotted:
When fetching the course score information, it wasn't fetching the user's specific enrollment in the course, so the interactor didn't have the user's score in the course available to it. I also changed it to use the `currentScore` value instead of the `computedFinalScore` value. Perhaps this is incorrect. But that's why we have code reviews :).

Screenshot 2025-04-04 at 10 07 33 AM

[ignore-commit-lint]

### DIFF
--- a/Horizon/Horizon/Sources/Features/Learn/CourseDetails/View/CourseDetailsView.swift
+++ b/Horizon/Horizon/Sources/Features/Learn/CourseDetails/View/CourseDetailsView.swift
@@ -47,7 +47,7 @@ struct CourseDetailsView: View {
             }
         }
         .refreshable {
-            await viewModel.refresh()
+            await viewModel.refresh(tab: Tabs.allCases[selectedTabIndex])
         }
     }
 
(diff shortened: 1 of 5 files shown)


szabinst (commented):
Can you provide a test account for me where the misbehaving is visible? I just checked 2 accounts and I see the same score as in web.

I have a feeling that we received this bug ticket when someone completed an assignment that got graded, expected the result to be displayed in the app but the app showed N/A because it wasn't refreshing properly.

reabbotted:
> Can you provide a test account for me where the misbehaving is visible? I just checked 2 accounts and I see the same score as in web.
> 
> I have a feeling that we received this bug ticket when someone completed an assignment that got graded, expected the result to be displayed in the app but the app showed N/A because it wasn't refreshing properly.

I'd been using my own account, reed.abbott@instructure. In the Basic Biology Certification, I went into the course and set grades for assignments and they were not appearing in iOS. But my understanding for where that data should come from is pretty rough.

szabinst (approved):

PR #3294, opened 2025-04-07
Student - NewQuizzes submission detail screen has balloon image messing up the layout

ndrsszsz:
## NewQuizzes submission detail screen has balloon image messing up the layout

refs: MBL-18735
affects: Student
release note: None
test plan: See ticket.

The issue was cause by the space for the attempt picker. When there's no attempt (no submission) then the attempt picker is hidden, so when the submission detail screen is opened and the accessibility font size is set to the largest then the top of the balloon panda from the empty view appears in that "window" where the attempt picker should be (if there were any submission). So the solution was to set the constraints of the view depending on the attempt picker (which depends on the submission). So when there is no attempt picker (no submission) then the top constraint of the content view is set to match the top constraint of the safe area (by default it's set to the bottom constraint of the attempt picker).

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [x] Tested on tablet

### DIFF
--- a/Student/Student/Submissions/SubmissionDetails/SubmissionDetailsViewController.swift
+++ b/Student/Student/Submissions/SubmissionDetails/SubmissionDetailsViewController.swift
@@ -46,6 +46,8 @@ class SubmissionDetailsViewController: ScreenViewTrackableViewController, Submis
     @IBOutlet weak var emptyView: SubmissionDetailsEmptyView?
     @IBOutlet weak var lockedEmptyView: SubmissionDetailsLockedEmptyView?
     @IBOutlet weak var attemptPicker: SubmissionAttemptPickerView?
+    @IBOutlet weak var dividerView: DividerView?
+    @IBOutlet weak var contentViewSafeAreaConstraint: NSLayoutConstraint?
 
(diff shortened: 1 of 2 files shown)


vargaat (approved):
QA+1

Student/Student/Submissions/SubmissionDetails/SubmissionDetailsViewController.swift
```diff
@@ -117,7 +119,7 @@ class SubmissionDetailsViewController: ScreenViewTrackableViewController, Submis
         guard let attemptPicker,
               let currentSubmission,
               let currentAttemptDate = currentSubmission.submittedAt
-        else { return }
+        else { return setConstraintsForNoSubmission() }
```
rh12, line 122: 1. Although this technically works since that method returns void, please first call the method and return sepatrately after that.
2. `attemptPicker?.isHidden` depends on `isSubmitted`, not this guard. Shouldn't we also handle that case? The originial logic is strange here, but nevertheless we should probably cover that too.
3. I think we should reset the proper constraints when this method is called again and there are actual attempts (i.e.: whenever `reload()` is called).

rh12 (changes requested):

rh12 (approved):
QA+1

PR #3348, opened 2025-05-06
SpeedGrader comments

rh12:
refs: [MBL-18809](https://instructure.atlassian.net/browse/MBL-18809)
affects: Teacher
release note: none

## What's changed
- Updated SpeedGrader comment list to match new design (see specific Figma link in ticket)
- Extracted related code to viewmodels, interactor, assembly, etc.

### Chores
- Added `SwiftUI.Image.size(_ size:, paddedTo boundingSize:)` method 
  - this allows us to more easily follow bounding boxes in Figma
- Added convenience method `sinkFailureOrValue()`
- Extracted various `MockError`s to a central place

## What's not changed
- Please see ticket for detailed scope first
- New comment is displayed as Sent instantly, regardless of the API response.
  - It was like this before, it probably needs a rework.
  - Even though I've created a `SubmissionCommentsInteractor` I didn't rework the create comment methods, just used them as is for now.
  - I haven't added tests for these
- Attempts are still in the comments, their design hasn't changed yet.
- The new Submission Comment related code is still in Teacher app, even though they could reused later in S-tudent app

## Screenshots
<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist
- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Teacher/Teacher/SpeedGrader/Comments/View/SubmissionCommentListCell.swift
+++ b/Teacher/Teacher/SpeedGrader/Comments/View/SubmissionCommentListCell.swift
@@ -19,6 +19,16 @@
 import SwiftUI
 import Core
 
+private enum Size {
+    static let verticalSpacing: CGFloat = 2
+    static let horizontalSpacing: CGFloat = 12
+    static let commentBubblePadding: CGFloat = 8
+    static let commentBubbleCorner: CGFloat = 16
+    static let attachmentSpacing: CGFloat = 4
+    static let attachmentCorner: CGFloat = 4
+    static let attachmentIcon: CGFloat = 18
+}
+
 struct SubmissionCommentListCell: View {
     @Environment(\.appEnvironment) var env
     @Environment(\.viewController) var controller
@@ -30,7 +40,7 @@ struct SubmissionCommentListCell: View {
 
     var body: some View {
         if viewModel.author.isCurrentUser {
-            VStack(alignment: .trailing, spacing: 2) {
+            VStack(alignment: .trailing, spacing: Size.verticalSpacing) {
                 header
                 commentView
             }
@@ -39,10 +49,10 @@ struct SubmissionCommentListCell: View {
             .padding(.top, 2)
             .padding(.bottom, 8)
         } else {
-            HStack(alignment: .top, spacing: 12) {
+            HStack(alignment: .top, spacing: Size.horizontalSpacing) {
                 avatar
                     .accessibilityHidden(true)
-                VStack(alignment: .leading, spacing: 2) {
+                VStack(alignment: .leading, spacing: Size.verticalSpacing) {
                     header
                     commentView
                 }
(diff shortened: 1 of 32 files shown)


inst-danger:

</details>

suhaibabsi-inst (changes requested):
Tested on iPhone XS, iOS 18.3.0

When tapping on textfield to input comment, the textfield scrolls up making it harder to follow on what's written before sending. See recording below:

https://github.com/user-attachments/assets/06686e6b-79ad-4e0f-85cc-6083215521bf

**_A minor issue:_**
With student comment that includes a lot of empty lines in between, the new design might get a bit confusing to figure things out. See below:

Simulator Screenshot - iPhone 16 Pro - 2025-05-08 at 23 22 58

Simulator Screenshot - iPhone 16 Pro - 2025-05-08 at 23 24 58

rh12:
@suhaibabsi-inst 
> When tapping on textfield to input comment, the textfield scrolls up making it harder to follow on what's written before sending. See recording below:

Nice find! I've seen this in simulator only for some reason and I thought it is because I used the actual keyboard instead of the simulator one. But it turns out it is a bug before the PR as well, but the comment library hid it before.
The same issue was there for other keyboard handlings in SpeedGrader.
It is now fixed.

> **_A minor issue:_** With student comment that includes a lot of empty lines in between, the new design might get a bit confusing to figure things out. See below:

Personally I agree with having bubbles as well, although I don't think many empty lines are a common occurrence in a comment. Since the design is set for now, I would not modify this, but feel free to bring this up in slack.

suhaibabsi-inst (approved):
QA +1

Thanks @rh12 !

ndrsszsz (dismissed):
QA +1
However, I agree with Suhaib on the design having a bad UX on some cases.

Teacher/Teacher/SpeedGrader/MainLayout/ViewModel/SubmissionGraderViewModel.swift
```diff
@@ -19,36 +19,58 @@
...
     @Published private(set) var file: File?
     @Published private(set) var fileID: String?
     @Published private(set) var fileTabTitle: String = ""
+    @Published private(set) var contextColor = Color(Brand.shared.primary)
     let assignment: Assignment
     let submission: Submission
 
+    // sub-viewmodels
+    @Published private(set) var studentAnnotationViewModel: StudentAnnotationSubmissionViewerViewModel
+    @Published private(set) var commentListViewModel: SubmissionCommentListViewModel
```
vargaat, line 42: We can safely remove the `@Published` annotations from here.

Teacher/Teacher/SpeedGrader/Comments/View/SubmissionCommentListView.swift
```diff
@@ -32,63 +30,53 @@ struct SubmissionCommentListView: View {
...
                             .frame(minWidth: geometry.size.width, minHeight: geometry.size.height - 40)
                     }
                 }
                     .background(Color.backgroundLightest)
-                    .scaleEffect(y: viewModel.state.isData ? -1 : 1)
-                Divider()
+                    .scaleEffect(y: viewModel.state == .data ? -1 : 1)
                 switch showRecorder {
                 case .audio:
+                    Divider()
```
vargaat, line 79: Can we use `InstUI.Divider()` here? Same question for the other divider changes in the PR.

rh12: These inline recorders will be replaced later to full screen ones as in inbox, but good call, replaced the dividers

Teacher/Teacher/SpeedGrader/Comments/View/SubmissionCommentListCell.swift
```diff
@@ -19,217 +19,259 @@
 import SwiftUI
 import Core
 
+private enum Size {
```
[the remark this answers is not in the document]
rh12: As discussed, this will stay at the top as it have constants for helper types as well

rh12:
Refactored `scaledSize()` method after discussion with @vargaat

rh12:
Waiting for UX approval, moving to draft for now

suhaibabsi-inst (commented):
Tested on iPhone XS, iOS 18.3.1

I found that there is a quite noticeable [hitch](https://developer.apple.com/documentation/xcode/understanding-hitches-in-your-app) when showing and dismiss keyboard, See below recording.

https://github.com/user-attachments/assets/eff6e438-ad17-4677-a6af-f61eaa1efa16

Moreover, design of textfield on Figma file is different from that developed, where buttons of attachment and quick responses is laid below the text input itself.

Screenshot 2025-05-27 at 4 18 53 PM

Also, there a white space between the very bottom of the screen and the bottom of screen safeArea, which should be filled with the same color of the input back. The same way as in Tab and Tool Bars.

IMG_2258

rh12:
@suhaibabsi-inst 
> I found that there is a quite noticeable [hitch](https://developer.apple.com/documentation/xcode/understanding-hitches-in-your-app) when showing and dismiss keyboard, See below recording.

The video doesn't load, could you please upload it again?
If this is about the issue when on iPads in portrait mode the textfield focus reloads the screen that is being fix by Attila below

> Moreover, design of textfield on Figma file is different from that developed, where buttons of attachment and quick responses is laid below the text input itself.

Yes, it had been stated that we use the old design for now, because the intended design waits for a11y approval.

> Also, there a white space between the very bottom of the screen and the bottom of screen safeArea, which should be filled with the same color of the input back. The same way as in Tab and Tool Bars.

Yes, it's been there before, as well. But since the old design will go away I didn't fix this one.

vargaat (approved):
QA+1, Let's to some optimization at the end when we delivered all changes.

rh12:
Thanks @vargaat for the rotation fix! It works for me on iPhone @ 18 and iPad @ 17

vargaat (approved):

PR #3359, opened 2025-05-08
[Horizon] Bringing back the inbox and skillspace

reabbotted:
[ignore-commit-lint]


inst-danger:

</details>

szabinst (commented):
I think we are missing a back button when you open Inbox and we should probably add pull to refresh to webViews. I'm fine doing the latter in a separate PR.

szabinst (approved):
Thanks for the web change, checked out that code too, looks good.

PR #3363, opened 2025-05-12
Fixed issue of Discussions being inaccessible for cross-shared users

suhaibabsi-inst:
refs: MBL-18802
affects: Student, Teacher
release note: Fixed issue of Discussions being inaccessible for cross-shared users.

## Test Plan

See ticket's description. Also, we need to test edit & create pages for discussions and announcements.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [x] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
+++ b/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
@@ -127,23 +127,21 @@ extension AppEnvironment {
 
 // MARK: - Utils
 
-public struct RoutePath {
-    let value: String
-    let env: AppEnvironment
+extension String {
 
-    /// This method returns the correct path using env `baseURL`
-    /// if it was different from the one used on `AppEnvironment.shared.api`
-    func correctedPath() -> String {
-        guard env.api.baseURL != AppEnvironment.shared.api.baseURL else {
-            return value
-        }
-        return env.api.baseURL.appending(path: value).absoluteString
+    public func asRoute(in env: AppEnvironment) -> String {
+        // Skip for full valid URLs
+        if isFullURLString { return self }
+
+        // Skip when baseURL equals to the shared one.
+        if env.api.baseURL == AppEnvironment.shared.api.baseURL { return self }
+
+        return env.api.baseURL.appending(path: self).absoluteString
     }
-}
 
-public extension String {
-    func asRoutePath(in env: AppEnvironment) -> RoutePath {
-        RoutePath(value: self, env: env)
+    private var isFullURLString: Bool {
+        guard let url = URL(string: self) else { return false }
+        return url.host() != nil
     }
 }
 
--- a/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
+++ b/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
@@ -127,6 +127,26 @@ extension AppEnvironment {
 
 // MARK: - Utils
 
+public struct RoutePath {
+    let value: String
+    let env: AppEnvironment
+
+    /// This method returns the correct path using env `baseURL`
+    /// if it was different from the one used on `AppEnvironment.shared.api`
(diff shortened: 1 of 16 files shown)


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Fixed issue of Discussions being inaccessible for cross-shared users.

#### Affected Apps: Student

[MBL-18802](https://instructure.atlassian.net/browse/MBL-18802)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d34d162904251ec2a4dea7eb98a843108307c08b
</p>

inst-danger:

</details>

inst-danger:

</details>

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
```diff
@@ -127,6 +127,26 @@ extension AppEnvironment {
 
 // MARK: - Utils
 
+public struct RoutePath {
```
vargaat, line 130: My feeling is that this `RoutePath` struct gives the solution a twist but the abstraction itself provides no added value as opposed to using strings directly. I would move the logic from here to a string extension and use that directly:
```
public extension String {
    func asRoute(in env: AppEnvironment) -> String {
        guard env.api.baseURL != AppEnvironment.shared.api.baseURL else {
            return value
        }
        return env.api.baseURL.appending(path: value).absoluteString
    }
}
```
This way no modification would be needed in the router.

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
```diff
@@ -127,6 +127,26 @@ extension AppEnvironment {
...
+    let value: String
+    let env: AppEnvironment
+
+    /// This method returns the correct path using env `baseURL`
+    /// if it was different from the one used on `AppEnvironment.shared.api`
+    func correctedPath() -> String {
+        guard env.api.baseURL != AppEnvironment.shared.api.baseURL else {
+            return value
+        }
+        return env.api.baseURL.appending(path: value).absoluteString
```
vargaat, line 140: Since we are working with strings can we add some logic if the input is already a fully qualified url and not just a path? Passing a full url here will definitely screw things up. Also, we could reinforce this solution with some unit tests.

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironmentOverride.swift
```diff
@@ -127,6 +127,26 @@ extension AppEnvironment {
 
 // MARK: - Utils
 
+public struct RoutePath {
+    let value: String
+    let env: AppEnvironment
+
+    /// This method returns the correct path using env `baseURL`
+    /// if it was different from the one used on `AppEnvironment.shared.api`
+    func correctedPath() -> String {
```
vargaat, line 136: Maybe a name called `asRoute` would be better since this method can return either a fully qualified URL or just a url path but its name suggests that it always return a path.

vargaat (changes requested):
These shard issues are now just finger exercises for you! :) I had some remarks on the business logic though.

vargaat (changes requested):
Thank you for the changes, looks cleaner now. When I tried to test the fix I got a 401 failure while loading the discussions list, could you please take a look?

vargaat (approved):
QA+1, Logging in with JTT to the hernandez domain successfully loaded the discussion. Let's disregard the act as user error for now as it wasn't working properly before this PR and pick it up when it comes back .

rh12 (approved):
Code + 1, thanks for the changes/fixes!

Approving, as I don't think there is any point for multiple devs to QA for this kind of issue.

PR #3366, opened 2025-05-13
Improve course fetching

szabinst:
This PR introduces a number of changes. 

## Additions/Changes 
- `HCourse` and `CDCourse` is the main course object from now on

- Created `DashboardInteractor` with two functions:
    - `getAndObserveCoursesWithoutModules` - gets a list of all courses and sets up a listener to refresh courses whenever a new module item gets refreshed in the background
        - it will ignore the cache if **pull to refresh triggers it** OR  a **module item gets completed** so the the course will show the correct progress 
    - `refreshModuleItemsUponCompletions` - listens for module item changes and fires a new request to update 1 specific item

- Modifies `GetCoursesInteractorLive`:
    - `getCourseWithModules` - gets a course, downloads modules and module items and keeps listening for database changes so it can get notified when `refreshModuleItemsUponCompletions` refreshes an item in the background
    - `getCoursesWithoutModules` - gets a list of all courses; does not include modules

- Added `GetLearnCoursesInteractor` as a Learn tab specific interactor that will get a list of all courses saved as `CDLearnCourse` in Core Data. The graphQL request only includes 3 properties so we can minimize the payload. 

## Fixes
- When fetching a course from graphQL we recently added modulesConnection in case the incompleteModulesConnection is empty. This was downloading all the modules so I limited it the `first: 1`
- Rewrote the logic how we handle incomplete module items in `CDCourse`.


inst-danger:

</details>

Ahmed-Naguib93 (approved):
Thanks for great improvement

PR #3382, opened 2025-05-19
[Horizon] Removing Trailing Commas

reabbotted:


armstrongnate (approved):

ktheory:
How can we catch these linter errors before they're committed?

szabinst:
@ktheory For some time we had a pre-commit hook but it was slow. Xcode should warn you at build time about lint errors, that's usually enough. 

We can however, revisit adding it back as a hook. It might have improved since then.

PR #3407, opened 2025-05-27
[Horizon] GraphQL Error

reabbotted:
We were hitting a limiting error in GraphQL for the number of type aliases in the query. The solution here is to remove a bunch of those aliases.

[ignore-commit-lint]


inst-danger:

</details>

PR #3409, opened 2025-05-28
Add paddings to SubmissionViewer

petkybenedek:
refs: [MBL-18849](https://instructure.atlassian.net/browse/MBL-18849)
affects: Teacher
release note: none

test plan:
 - Check if the text on the SubbmissionViewer screen has padding.

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Teacher/Teacher/SpeedGrader/SubmissionRenderer/View/SubmissionViewer.swift
+++ b/Teacher/Teacher/SpeedGrader/SubmissionRenderer/View/SubmissionViewer.swift
@@ -27,6 +27,7 @@ struct SubmissionViewer: View {
 
     @Environment(\.appEnvironment) var env
     @Environment(\.viewController) var controller
+    @Environment(\.dynamicTypeSize) var dynamicTypeSize
     @ObservedObject private var studentAnnotationViewModel: StudentAnnotationSubmissionViewerViewModel
 
     public init(assignment: Assignment, submission: Submission, fileID: String?, studentAnnotationViewModel: StudentAnnotationSubmissionViewerViewModel, handleRefresh: (() -> Void)?) {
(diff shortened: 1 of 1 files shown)


inst-danger:

</details>

Teacher/Teacher/SpeedGrader/SubmissionRenderer/View/SubmissionViewer.swift
```diff
@@ -65,7 +66,9 @@ struct SubmissionViewer: View {
                     Text("This student's responses are hidden because this assignment is anonymous.", bundle: .teacher)
                     Spacer()
                 }
-                .font(.regular16).foregroundColor(.textDarkest)
+                .font(.regular16)
+                .padding(InstUI.Styles.Padding.standard.rawValue)
```
rh12: We could use the standard value directly:
`.paddingStyle(.all, .standard)` or `.paddingStyle(.horizontal, .standard)`

(For some reason `.paddingStyle(.standard)` doesn't work, this could be fixed later)

Same for other usage below

petkybenedek: Swift will not figure out which parameter was specified between multiple, non-named parameters with default values. Here's a short example without type inference:

```
func testFunction(_ message: String = "", _ statusCode: Int = 200) {
    print("Test Function Called: \(message) with status code: \(statusCode)")
}

testFunction(200)    // Cannot convert value of type 'Int' to expected argument type 'String'
testFunction()
testFunction("Open in Safari")
testFunction("Open in Safari", 200)
```

I fixed it, `.paddingStyle(.standard)` can be used now.

rh12 (approved):
QA + 1
Thanks for cleaning up the paddings method!

balintbartok (approved):
QA + 1

PR #3417, opened 2025-06-02
Horizon Core Changes 

szabinst:
Adds: 
- new core data entities
- new graphql requests
- new use cases
- tests to cover these areas

These changes should not affect any of the apps. 


inst-danger:

</details>

inst-danger:

</details>

inst-danger:

</details>

Core/Core/Features/Modules/MarkModuleItemDone.swift
```diff
@@ -0,0 +1,76 @@
...
+                )
+            }
+            completionHandler(response, urlResponse, error)
+        }
+    }
+    public func write(
+        response: APINoContent?,
+        urlResponse: URLResponse?,
+        to client: NSManagedObjectContext
+    ) { }
```
vargaat, line 75: Don't we want to update the associated ModuleItem in CoreData? I think that would be the expected for such a UseCase.

Core/Core/Features/Modules/MarkModuleItemRead.swift
```diff
@@ -0,0 +1,73 @@
...
+            }
+            completionHandler(response, urlResponse, error)
+        }
+    }
+
+    public func write(
+        response: APINoContent?,
+        urlResponse: URLResponse?,
+        to client: NSManagedObjectContext
+    ) { }
```
vargaat, line 72: Don't we want to update the associated ModuleItem in CoreData? I think that would be the expected for such a UseCase.

Core/Core/Features/Notebook/Model/CDNotebookNote.swift
```diff
@@ -0,0 +1,58 @@
...
+extension String? {
+    public var deserializeLabels: [String]? {
+        self?.deserializeLabels
+    }
+}
+
+extension Array where Element == String {
+    public var serializeLabels: String? {
+        sorted().joined(separator: ";")
+    }
```
vargaat: It's strange for me that these specialized methods are just added as public extensions available for everyone. If these are connected to CDNoteBookNote than we should add these to that class.

Core/Core/Features/Modules/MarkModuleItemDone.swift
szabinst: The API only returns an `"message": "OK"`.

Core/Core/Features/Modules/MarkModuleItemRead.swift
szabinst: I haven't checked this specifically, but I'm pretty sure it's the same as above.

Core/CoreTests/Features/Courses/CourseProgression/Model/CDCourseTests.swift
```diff
@@ -0,0 +1,177 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+@testable import Core
+import XCTest
+
+class CDCourseTests: CoreTestCase {
```
rh12: Could you please add the `H` prefix to the tests as well? (filename & class)

Same for
- all in this folder
- in Notebook/Model
- CDScoresAssignmentTests

Core/CoreTests/Features/Courses/CourseProgression/Model/CDCourseTests.swift
```diff
@@ -0,0 +1,177 @@
...
+                name: "Completed Course",
+                account: GetHCoursesProgressionResponse.AccountModel(name: "Completed Account"),
+                imageUrl: "https://example.com/completed.png",
+                syllabusBody: "Completed syllabus",
+                usersConnection: GetHCoursesProgressionResponse.UsersConnection(
+                    nodes: [
+                        GetHCoursesProgressionResponse.NodeModel(
+                            courseProgression: GetHCoursesProgressionResponse.CourseProgression(
+                                requirements: GetHCoursesProgressionResponse.Requirements(completionPercentage: 100.0),
+                                incompleteModulesConnection: GetHCoursesProgressionResponse.IncompleteModulesConnection(nodes: [])
```
rh12: I didn't check, but would it be possible to use `.init(...)` instead all these long typenames?
(I assume the topmost init needs the explicit type, but the others do not)
It's a bit hard to read now IMHO

Same in other places

Core/CoreTests/Features/Courses/CourseProgression/Model/GetCoursesProgressionUseCaseTests.swift
```diff
@@ -0,0 +1,95 @@
...
+            id: "enroll_1",
+            course: GetHCoursesProgressionResponse.CourseModel(
+                id: "course_1",
+                name: "Test Course",
+                account: nil,
+                imageUrl: nil,
+                syllabusBody: nil,
+                usersConnection: nil,
+                modulesConnection: nil
+            )
```
rh12: Would it make sense to create (generate?) the usual `.make()` methods for these with minimal default values?
(nil if optional, "" if String, 0 if Int, etc.)
I see those are not created for any of these models (maybe intentionally).

Core/CoreTests/Features/Courses/CourseProgression/Model/CDCourseTests.swift
szabinst: I already renamed some of those classes but looks like xcode didn't rename the files themselves. Fixed it now!

Core/CoreTests/Features/Courses/CourseProgression/Model/CDCourseTests.swift
szabinst: The first iteration was like that, but it was harder to read and process how this graphql structure looks like so I opted for the style above.

Core/CoreTests/Features/Courses/CourseProgression/Model/GetCoursesProgressionUseCaseTests.swift
szabinst: I personally find the `.make()` method almost unusable because of how Xcode handles it. Usually, it takes more time to try to find the init method by typing a `.` to initiate the completion handler, then cherry pick the properties where you want to actually pass a value than simply having an initializer where you fill out every parameter even if it's nil.

rh12: I agree that current Xcode handles parameters in a crappy way... autocompletion used to work properly, but now it's just pain.
IMHO the advantage of the `.make()` methods is the much less noise at the call site. So not writing less, but reading less.
With all that said I'm fine with this as it is.

rh12 (approved):

Core/Core/Features/Notebook/Model/CDNotebookNote.swift
szabinst: Added them as a function to the entity itself.

vargaat (approved):

PR #3428, opened 2025-06-06
SpeedGrader improvements

vargaat:
### Bug Fixes
- SpeedGrader header didn't show `Graded` as status even if on the previous screen you tapped on a student with graded status. The status display logics are now in sync.
- There was an issue on iOS 17 devices in landscape mode when the comments tab was selected: the invisible grades tab blocked user interaction on the submission area/attempt picker.
- There was an untranslated label for on paper submissions. The translation was available in the core bundle so I switched to use it.<br/>

### Performance Improvements
I noticed the poor performance in SpeedGrader landscape mode when you resize the right column so here are a few improvements that make the animation smoother but does not completely solve the issue.
- One thing that was definitely slow is the formatted string in the `User.displayName` method so I cached the result of this method instead of calculating it in the view's body. The submission header view and the comment cell was affected.
- In `SubmissionHeaderView` we saved the container's size to update the profile image's size when the header height changes. Only the height matters so updating this State property when only the width changes could have caused unnecessary redraws.
- I extracted a few properties to `SubmissionHeaderViewModel` that were calculated during body renders but I don't think these increase the performance significantly but since I had the view model because of the `User.displayName` optimization it was straightforward to move these there.

refs: MBL-18832
affects: Teacher
release note: none

test plan:
- Start SpeedGrader in landscape mode.
- Resize the right column by dragging it.
- Performance should be acceptable.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Teacher/Teacher/Assignments/AssignmentDetails/Views/TeacherSubmissionBreakdownView.swift
+++ b/Teacher/Teacher/Assignments/AssignmentDetails/Views/TeacherSubmissionBreakdownView.swift
@@ -76,7 +76,7 @@ struct TeacherSubmissionBreakdownView<ViewModel: SubmissionBreakdownViewModelPro
                             color: viewModel.color
                         )
                         Text(String.localizedStringWithFormat(
-                            String(localized: "there_are_d_assignees_without_grades", bundle: .teacher),
+                            String(localized: "there_are_d_assignees_without_grades", bundle: .core),
                             viewModel.ungraded + viewModel.unsubmitted
                         ))
                         .font(.regular14).foregroundColor(.textDarkest)


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>This pull request will not generate a release note.</td>
    </tr>
  </tbody>
</table>


#### Affected Apps: Teacher

[MBL-18832](https://instructure.atlassian.net/browse/MBL-18832)
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8d39df277e5e453ae7d7bff62f04172e74a7e0a2
</p>

rh12 (approved):
QA + 1

rh12 (approved):
QA + 1

PR #3433, opened 2025-06-10
Fix the issue of cross-shared user unable to preview files added to Modules

suhaibabsi-inst:
refs: MBL-18871
affects: Student
release note: Fixed the issue where cross-shared user not being able to preview files when added to Modules.

## Test Plan
See [ticket's](https://instructure.atlassian.net/browse/MBL-18871) description.

#### :warning: Known issue:
~~When using "act as user" as login methodology, modules won't load as expected and user won't be able to preview files.
This issue is caused by an error is returned from the API when `as_user_id` query param is passed along. (Not sure why that happens?)~~

Just tested this issue today, seems to work normally, same way as the ordinary login way. See recording below:

https://github.com/user-attachments/assets/1e880c00-58fd-45b9-a228-29e269ef4a02

## Checklist

- [ ] A11y checked
- [x] Tested on phone
- [x] Tested on tablet
- [x] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Modules/ModuleList/ModuleListViewController.swift
+++ b/Core/Core/Features/Modules/ModuleList/ModuleListViewController.swift
@@ -29,7 +29,7 @@ public final class ModuleListViewController: ScreenViewTrackableViewController, 
     @IBOutlet weak var tableView: UITableView!
     public let titleSubtitleView = TitleSubtitleView.create()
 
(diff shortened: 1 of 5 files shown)


rh12 (approved):
QA + 1 (tested with normal login only)

vargaat (approved):
QA+1

PR #3448, opened 2025-06-17
[Horizon] Add profile disable name fields

Ahmed-Naguib93:
### What is new?
1- Fixed the issue where the mail was not appearing in PD
2- https://instructure.atlassian.net/browse/CLX-1514

### DIFF
--- a/Core/Core/Features/People/APIUser.swift
+++ b/Core/Core/Features/People/APIUser.swift
@@ -151,12 +151,10 @@ public struct APIProfile: Codable, Equatable {
     public struct Permissions: Codable, Equatable {
         public let canUpdateName: Bool?
         public let canUpdateAvatar: Bool?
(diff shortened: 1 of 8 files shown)


Core/Core/Features/People/APIUser.swift
```diff
@@ -146,6 +146,19 @@ public struct APIProfile: Codable, Equatable {
     public let uuid: String?
     public let account_uuid: String?
     public let time_zone: String?
+    public let permissions: APIProfile.Permissions?
+
+    public struct Permissions: Codable, Equatable {
+        public let canUpdateName: Bool?
+        public let canUpdateAvatar: Bool?
+        public let limitParentAppWebAccess: Bool?
```
szabinst: I think this is not used.

PR #3450, opened 2025-06-19
SpeedGrader - Add submission statuses

vargaat:
- Grade statuses are fetched from GraphQL and modifications of these are also uploaded via GraphQL because the REST API here is a bit messy.
- I extended our REST entities to be able to determine the grade status. It can be constructed from these 3 properties: `latePolicyStatus`, `custom_grade_status_id`, `excused`.
- We have a `LatePolicyStatus` enum for REST API in Core but it cannot be used directly for GraphQL because it doesn't include the `excused` status (since it's a separate Bool in REST).
- I fixed the issue of the status in the header not updating while changing the grade status only after re-launching the student's submission in SpeedGrader.

### Chores
- Added accessibility id to InstUI menu options so they can be targeted directly.
- Improved `setNetworkState*` methods to include `file` and `line` parameters so when they fail you know which line to comment out. 😬

refs: [MBL-18906](https://instructure.atlassian.net/browse/MBL-18906)
affects: Teacher
release note: Added grade status picker to SpeedGrader.

test plan:
- Enabled the extended grade status feature flag.
- Setup late penalty on your account.
- Add at lest a custom grade status.
- Start SpeedGrader.
- Select all status options, check if they properly sync with the web and vice versa.
- Change attempts, check if status reflects the current attempt.
- Add an excused grade, check if status changes to excused.
- Change late days and check if late penalty is re-calculated properly and in sync with web.

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [x] Approve from product

[MBL-18906]: https://instructure.atlassian.net/browse/MBL-18906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


#### Release Note: 
Added grade status picker to SpeedGrader.

#### Affected Apps: Teacher

[MBL-18906](https://instructure.atlassian.net/browse/MBL-18906)

**Canvas iOS** | 91.11% | 91.12% | -0%
Student/Student/WidgetRouter.swift | 0% | 0% | 0%
Student/Student/WidgetRouter+TodoWidget.swift | 0% | 0% | 0%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c582f17c58ebd6efd309fc7dbdda9f8bbd0e61af
</p>

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+        userId: String,
+        submissionId: String,
+        attempt: Int,
+        interactor: GradeStatusInteractor
+    ) {
+        self.interactor = interactor
+        self.submissionId = submissionId
+        self.userId = userId
+        options = interactor.gradeStatuses
+            .map { OptionItem(id: $0.id, title: $0.name, accessibilityId: "SpeedGrader.GradeStatusMenuItem.\($0.id)") }
```
rh12: In existing cases, the OptionItem's `id` is used as an `accessibilityId`. The prefix (= group) part of the a11yId is provided from outside.
I propose for `PickerMenu` itself to have an `identifierGroup` property, just like `SingleSelectionView` for example.
(then we can remove the newly introduced, somewhat redundant `OptionItem.accessibilityId`)

vargaat: Implemented, good idea.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/Model/GradeStatusInteractor.swift
```diff
@@ -0,0 +1,168 @@
...
+            .catch { _ in Publishers.typedEmpty() }
+            .eraseToAnyPublisher()
+    }
+
+    func updateLateDays(
+        submissionId: String,
+        userId: String,
+        daysLate: Int
+    ) -> AnyPublisher<Void, Error> {
+        let lateSeconds = daysLate * 86400
```
suhaibabsi-inst:
```suggestion
        let lateSeconds = daysLate * 24 * 60 * 60
```

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
```diff
@@ -0,0 +1,107 @@
...
+                        viewModel.didChangeLateDaysValue.send(newDaysLateValue)
+                    }
+                )
+            }
+        }
+        .animation(.smooth, value: viewModel.isShowingDaysLateSection)
+    }
+
+    private var cell: some View {
+        HStack(spacing: InstUI.Styles.Padding.cellAccessoryPadding.rawValue) {
```
suhaibabsi-inst, line 41: Would be nice if we have an extension method on `CGFloat` that takes "InstUI.Styles.Padding" as an argument, in order to avoid accessing `rawValue`.

```swift
extension CGFloat {
      static func instuiPadding(_ padding: InstUI.Styles.Padding) -> CGFloat { padding.rawValue }
}
```
Then, calling that like:
```swift
HStack(spacing: .instuiPadding(.cellAccessoryPadding))
```
CC: @rh12

suhaibabsi-inst:
QA +1

I was able to match what is on iOS with that experience on the Web.

suhaibabsi-inst (dismissed):

Teacher/Teacher/Localizable.xcstrings
```diff
@@ -13574,6 +13608,9 @@
           }
         }
       }
+    },
+    "Due: %@" : {
```
rh12: For consistency we changed this to `Due %@` in the header, probably we should do it here as well and remove this variant

Teacher/Teacher/SpeedGrader/StudentPager/ViewModel/SpeedGraderViewModel.swift
```diff
@@ -185,11 +185,16 @@ extension SpeedGraderViewModel: PagesViewControllerDataSource {
...
                 latestSubmission: data.submissions[index],
                 contextColor: interactor.contextInfo.compactMap { $0?.courseColor }.eraseToAnyPublisher(),
+                gradeStatusInteractor: interactor.gradeStatusInteractor,
                 env: environment
             ),
             landscapeSplitLayoutViewModel: landscapeSplitLayoutViewModel,
             handleRefresh: { [weak self] in
-                self?.interactor.refreshSubmission(forUserId: data.submissions[index].userID)
+                guard let self else { return }
+                self.interactor
```
rh12: `self` could be removed after the guard (here and few lines below)

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/Model/GradeStatus.swift
```diff
@@ -0,0 +1,51 @@
...
+    let name: String
+    let isCustom: Bool
+
+    init(custom: APIGradeStatuses.CustomGradeStatus) {
+        self.id = custom.id
+        self.name = custom.name
+        self.isCustom = true
+    }
+
+    init(defaultName: String) {
```
rh12: Could we please explain these 2 inits in a comment, which is to be used when/why?
Please also mention that custom grades are already localized (or not) by the teacher who defines them.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/Model/GradeStatus.swift
```diff
@@ -0,0 +1,51 @@
...
+extension String {
+
+    internal var localizedGradeStatusName: String {
+        switch self {
+        case "late": String(localized: "Late", bundle: .teacher)
+        case "missing": String(localized: "Missing", bundle: .teacher)
+        case "excused": String(localized: "Excused", bundle: .teacher)
+        case "extended": String(localized: "Extended", bundle: .teacher)
+        case "none": String(localized: "None", bundle: .teacher)
+        default: capitalized
```
rh12: It's still a bit weird to me that we define our known cases in a switch here, and not in an enum or something.
Even if that enum never leaves this file, it would still be clearer IMHO. But I'm not gonna die on that hill :)

Could you please code comment a bit on each case, what they really mean? (Late vs Missing, whats Extended, None being the regular value)

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/Model/GradeStatusInteractor.swift
```diff
@@ -0,0 +1,168 @@
...
+                    )
+                else {
+                    return nil
+                }
+                let daysLate = Int(ceil(Double(submission.lateSeconds) / 86400.0))
+                let dueDate = submission.assignment?.dueAt
+                return (status, daysLate, dueDate)
+            }
+            .removeDuplicates { lhs, rhs in
+                lhs.0.id == rhs.0.id && lhs.1 == rhs.1 && lhs.2 == rhs.2
```
rh12: Could we maybe either define Equatable in GradeStatus like this? Or just simply compare all properties and do
`.removeDuplicates()`

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
```diff
@@ -0,0 +1,152 @@
...
+                    onEdit: { _ in }
+                )
+                .tint(Color.red)
+            }
+        }
+        .previewLayout(.sizeThatFits)
+    }
+}
+
+struct CoreHostingControllerPreview<Content: View>: UIViewControllerRepresentable {
```
rh12: This seems general enough to be somewhere in `CommonUI`.
If not, please make it private

Teacher/TeacherTests/SpeedGrader/Grading/GradeStatus/Model/GradeStatusInteractorTests.swift
```diff
@@ -0,0 +1,163 @@
...
+//
+
+import XCTest
+@testable import Core
+import Combine
+@testable import Teacher
+import TestsFoundation
+
+class GradeStatusInteractorTests: TeacherTestCase {
+    private var cancellables: Set<AnyCancellable> = []
```
rh12: nit
```suggestion
    private var subscriptions: Set<AnyCancellable> = []
```

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,249 @@
...
+import SwiftUI
+
+class GradeStatusViewModel: ObservableObject {
+    // MARK: - Outputs
+    @Published private(set) var selectedOption = OptionItem.none
+    @Published private(set) var isLoading = false
+    @Published private(set) var isShowingDaysLateSection = false
+    @Published private(set) var daysLate = ""
+    @Published private(set) var daysLateA11yLabel = ""
+    @Published private(set) var dueDate = ""
```
rh12: Could you please add the types, even if it's trivial?
(Especially for the optional `selectedOption`)

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,249 @@
...
+
+    private func uploadGradeStatus(
+        on publisher: PassthroughSubject<OptionItem, Never>
+    ) {
+        publisher
+            .compactMap { [weak self, interactor] selectedOption -> (OptionItem, GradeStatus)? in
+                guard
+                    let self,
+                    self.selectedOption != selectedOption, // If the user selects the same option, we don't need to do anything
+                    let selectedStatus = interactor.gradeStatuses.first(where: { $0.id == selectedOption.id })
```
rh12: If you conform `GradeStatus` to `OptionItemIdentifiable`, then you can use `interactor.gradeStatuses.element(for: selectedOption)`
Maybe it could simplify optionitem-model mappings elsewehere, too.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,249 @@
...
+
+private extension OptionItem {
+    var isLate: Bool {
+        id == LatePolicyStatus.late.rawValue
+    }
+
+    static var none: OptionItem {
+        OptionItem(
+            id: "none",
+            title: String(localized: "None", bundle: .teacher)
```
rh12: If possible, could we create the actual `GradeStatus(defaultName: "none")` instead and simply map that to optionutem as usual?

rh12 (commented):
Had some code findings, will review the new Views (and the related VM) during QA

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
```diff
@@ -0,0 +1,107 @@
...
+            if viewModel.isLoading {
+                ProgressView()
+                    .tint(nil)
+            } else {
+                Text(viewModel.selectedOption.title)
+                    .font(.regular14)
+                    .foregroundStyle(.tint)
+                Image.chevronDown
+                    .scaledIcon(size: 24)
+                    .foregroundStyle(.tint)
```
rh12: 1. Shouldn't tint apply automatically? We even need to neturalize it in ProgressView a bit above (if I understand correctly)
2. Figma doesn't use tint, it uses textdark here, and I think that's consistent with other cells (so question 1 is a bit theoretical)

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+        .frame(minHeight: 52)
+        .paddingStyle(.horizontal, .standard)
```
rh12: I'm not sure where the vertical padding is applied (it's nowhere and we just use the minheight above?)
I think we should use `.paddingStyle(set: .standardCell)` instead.
Same for the DaysLateView

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
```diff
@@ -0,0 +1,107 @@
...
+            if viewModel.isShowingDaysLateSection {
+                GradeStatusDaysLateView(
+                    daysLate: viewModel.daysLate,
+                    dueDate: viewModel.dueDate,
+                    isLoading: viewModel.isLoading,
+                    accessibilityLabel: viewModel.daysLateA11yLabel,
+                    accessibilityHint: viewModel.daysLateA11yHint,
+                    onEdit: { newDaysLateValue in
+                        viewModel.didChangeLateDaysValue.send(newDaysLateValue)
+                    }
```
rh12: would it make sense to pass in the viewmodel instead?

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,243 @@
...
+class GradeStatusViewModel: ObservableObject {
+    // MARK: - Outputs
+    @Published private(set) var selectedOption: OptionItem
+    @Published private(set) var isLoading: Bool = false
+    @Published private(set) var isShowingDaysLateSection: Bool = false
+    @Published private(set) var daysLate: String = ""
+    @Published private(set) var daysLateA11yLabel: String = ""
+    @Published private(set) var dueDate: String = ""
+    @Published var isShowingSaveFailedAlert: Bool = false
+    let daysLateA11yHint = String(localized: "Double-tap to change late days number.", bundle: .teacher)
```
rh12: Could you please add a line between these two for me? :)

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,243 @@
...
+                    return nil
+                }
+                self.isLoading = true
+                let oldOption = self.selectedOption
+                // We select the option now to make voiceover properly handle focus
+                // after the menu is dismissed. If the request fails we revert to the old value.
+                self.selectedOption = selectedOption
+
+                // If late is disabled we instantly hide the days late section to avoid
+                // inconsistencies of it being edited while the disable request is in progress.
```
rh12:
```suggestion
                // If late is not selected we instantly hide the days late section to avoid
                // inconsistencies of it being edited while the status change request is in progress.
```

I almost started to look for feature flags and settings on how to disable the "Late" option.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
```diff
@@ -0,0 +1,243 @@
...
+            .store(in: &subscriptions)
+    }
+
+    private static func announceSuccessfulSaveIfNecessary(
+        result: Result<Void, Error>,
+        oldOption: OptionItem
+    ) -> AnyPublisher<(Result<Void, Error>, OptionItem), Never> {
+        if result.isFailure {
+            return Just((result, oldOption))
+                .eraseToAnyPublisher()
```
rh12, line 228: Shouldn't we announce failure as well?

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
```diff
@@ -0,0 +1,140 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Core
+import SwiftUI
+
+struct GradeStatusDaysLateView: View {
```
rh12, line 22: Later on we could extract these kind of cells (this and the picker cell above it) similarly to `LabelValueCell`, and maybe rethink naming.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
```diff
@@ -0,0 +1,140 @@
...
+    @Environment(\.viewController) private var viewController
```
rh12, line 24: the view doesn't scale dynamically, we should probably add `@Environment(\.dynamicTypeSize) private var dynamicTypeSize`

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
```diff
@@ -0,0 +1,140 @@
...
+        .paddingStyle(.bottom, .cellBottom)
+        .paddingStyle(.horizontal, .standard)
+        .background(Color.backgroundLightest)
+        .accessibilityElement()
+        .accessibilityAddTraits(.isButton)
+        .accessibilityAction { presentNumberInputDialog() }
+        .accessibilityLabel(accessibilityLabel)
+        .accessibilityHint(accessibilityHint)
+        .accessibilityFocused($isA11yFocused)
+        .identifier("SpeedGrader.DaysLateButton")
```
rh12: I would suggest to move this one level outer, so both cell identifiers are defined on the same level

rh12 (commented):
QA + 1 (minus some findings below)

It's unclear to me what is the expectation on submission status vs grade status, in cases of Late, Missing, Excused (so the ones we have in both kind of statuses).
Currently Submission Status is not reflected automatically on Grade Status (we always start with "None", regardless of the submission status being e.g. "Missing")
On the other hand, changing the Grade Status via the new controls updates the submission status as expected (caveats below)
This needs more investigation (didn't thorouhgly test web behaviour) and either fix here to have a not-confusing behaviour, or in a followup, depending on product decisions.

Known somewhat unrelated issues:
- SG Header status is not updated upon grading or grade status change (it's updated on reopening SG or moving to another submission and back)
- Starting with a Submitted submission status, changing to Late grade status:
This updates the submission status to Late (as expected), but the Submission List will show the Late submission in "Other" section instead of "Submitted" (as it does for regular Late submissions)

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
vargaat: 1. Yup, I removed them.
2. Figma is outdated, the latest I know that we should color this as we color other editable fields in the new SpeedGrader design.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
vargaat: It was just the minHeight adding padding. I modified it to standardCell padding. It's now 2 points smaller then in figma but at least the paddings are consistent.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/ViewModel/GradeStatusViewModel.swift
vargaat: We pop up an error dialog, I think that's enough to draw attention.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
vargaat: Agree but only if this will be a common cell type and not just something made up only for SpeedGrader.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
vargaat: Added.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusDaysLateView.swift
vargaat: Moved.

Teacher/Teacher/SpeedGrader/Grading/GradeStatus/View/GradeStatusView.swift
vargaat: Makes sense, extracted.

vargaat:
> Found one last issue:
> I think the Status cell shouldn't act as a button, only the trailing part.
> One reason is that we have no dividers above/below, so one part of the big empty area activates the picker, and some pixels above or below doesn't.
> Another reason is that on iPads the picker menu shows up in horizontal center of the cell, not at the position of the trailing button.

Fixed it, nice improvement. I also pre-filled the alert textfield with the days late value so it's clearer what was the original value that we want to modify.

PR #3467, opened 2025-06-29
Fixed Consortia Root Teacher user access issue to Assignments

suhaibabsi-inst:
refs: MBL-18965
affects: Student, Teacher
release note: Fixed issue of Root account Teacher user unable to access Assignments.

Changes touched all other features that are incorporated into Assignment details & submission list screens, mainly: SpeedGrader, Send Message, Post Grades settings, hence the large number of changed files. 

Changes were shallow in nature, primarily making sure to pass the right `env` to views and model objects that use it.

## Test Plan

See [ticket's](https://instructure.atlassian.net/browse/MBL-18965) description.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
+++ b/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
@@ -63,6 +63,10 @@ open class AppEnvironment {
         self.logger = Logger.shared
     }
 
+    public var isShared: Bool {
+        api.baseURL == AppEnvironment.shared.api.baseURL
+    }
+
     /**
      - parameters:
         - isSilent: If this parameter is true then the method won't trigger a widget refresh
--- a/Core/Core/Features/Assignments/AssignmentDetails/AssignmentSubmissionBreakdownViewModel.swift
+++ b/Core/Core/Features/Assignments/AssignmentDetails/AssignmentSubmissionBreakdownViewModel.swift
@@ -41,15 +41,15 @@ public class AssignmentSubmissionBreakdownViewModel: SubmissionBreakdownViewMode
     private let courseID: String
     private let submissionTypes: [SubmissionType]
     private var summary: Store<GetSubmissionSummary>
-    private var submissionsPath: String { "courses/\(courseID)/assignments/\(assignmentID)/submissions" }
+    private var submissionsPath: String { "/courses/\(courseID)/assignments/\(assignmentID)/submissions" }
 
-    public init(courseID: String, assignmentID: String, submissionTypes: [SubmissionType], color: UIColor? = nil) {
+    public init(env: AppEnvironment, courseID: String, assignmentID: String, submissionTypes: [SubmissionType], color: UIColor? = nil) {
         self.assignmentID = assignmentID
         self.courseID = courseID
         self.submissionTypes = submissionTypes
         self.color = color?.asColor ?? .accentColor
 
-        summary = AppEnvironment.shared.subscribe(GetSubmissionSummary(
+        summary = env.subscribe(GetSubmissionSummary(
             context: .course(courseID),
             assignmentID: assignmentID
         ))
--- a/Core/Core/Features/Assignments/AssignmentEditor/AssignmentAssigneeList.swift
+++ b/Core/Core/Features/Assignments/AssignmentEditor/AssignmentAssigneeList.swift
@@ -42,10 +42,10 @@ struct AssigmentAssigneeList: View {
         !isEveryoneMatching && sections.isEmpty && groups.isEmpty && students.isEmpty
     }
 
-    init(courseID: String, groupCategoryID: String?, selection: Binding<[Assignee]>) {
-        groups = AppEnvironment.shared.subscribe(GetGroupsInCategory(groupCategoryID))
-        sections = AppEnvironment.shared.subscribe(GetCourseSections(courseID: courseID))
-        students = AppEnvironment.shared.subscribe(GetContextUsers(context: .course(courseID), type: .student))
+    init(env: AppEnvironment, courseID: String, groupCategoryID: String?, selection: Binding<[Assignee]>) {
+        groups = env.subscribe(GetGroupsInCategory(groupCategoryID))
+        sections = env.subscribe(GetCourseSections(courseID: courseID))
+        students = env.subscribe(GetContextUsers(context: .course(courseID), type: .student))
         _selection = selection
     }
 
--- a/Core/Core/Features/Assignments/AssignmentEditor/AssignmentAssigneePicker.swift
+++ b/Core/Core/Features/Assignments/AssignmentEditor/AssignmentAssigneePicker.swift
@@ -40,16 +40,16 @@ struct AssignmentAssigneePicker: View {
         groups.state == .loading || sections.state == .loading || students.state == .loading
     }
 
-    init(courseID: String, groupCategoryID: String?, overrides: Binding<[Override]>, override: Override) {
+    init(env: AppEnvironment, courseID: String, groupCategoryID: String?, overrides: Binding<[Override]>, override: Override) {
         self.courseID = courseID
         self.groupCategoryID = groupCategoryID
         self.override = override
         self._overrides = overrides
         self._selection = State(initialValue: Assignee.from(override))
 
-        groups = AppEnvironment.shared.subscribe(GetGroupsInCategory(groupCategoryID))
-        sections = AppEnvironment.shared.subscribe(GetCourseSections(courseID: courseID))
-        students = AppEnvironment.shared.subscribe(GetContextUsers(context: .course(courseID), type: .student))
+        groups = env.subscribe(GetGroupsInCategory(groupCategoryID))
+        sections = env.subscribe(GetCourseSections(courseID: courseID))
+        students = env.subscribe(GetContextUsers(context: .course(courseID), type: .student))
     }
 
     var body: some View {
@@ -132,9 +132,18 @@ struct AssignmentAssigneePicker: View {
     func student(_ id: String) -> User? { students.first { $0.id == id } }
 
     func add() {
-        env.router.show(CoreHostingController(AssigmentAssigneeList(
-            courseID: courseID, groupCategoryID: groupCategoryID, selection: $selection
-        )), from: controller)
+        env.router.show(
+            CoreHostingController(
+                AssigmentAssigneeList(
+                    env: env,
+                    courseID: courseID,
+                    groupCategoryID: groupCategoryID,
+                    selection: $selection
+                ),
+                env: env
+            ),
+            from: controller
+        )
     }
 
     func save() {
--- a/Core/Core/Features/Inbox/Addressbook/AddressbookAssembly.swift
+++ b/Core/Core/Features/Inbox/Addressbook/AddressbookAssembly.swift
@@ -23,7 +23,7 @@ import UIKit
 public enum AddressBookAssembly {
 
     public static func makeAddressbookRecipientViewController(
-        env: AppEnvironment = .shared,
+        env: AppEnvironment,
         recipientContext: RecipientContext,
         roleName: String,
         recipients: [Recipient],
@@ -40,11 +40,11 @@ public enum AddressBookAssembly {
             selectedRecipients: selectedRecipients
         )
         let view = AddressbookRecipientView(model: viewModel)
-        return CoreHostingController(view)
+        return CoreHostingController(view, env: env)
     }
 
     public static func makeAddressbookRoleViewController(
-        env: AppEnvironment = .shared,
+        env: AppEnvironment,
         recipientContext: RecipientContext,
         teacherOnly: Bool,
         didSelectRecipient: PassthroughRelay<Recipient>,
@@ -61,7 +61,7 @@ public enum AddressBookAssembly {
             selectedRecipients: selectedRecipients
         )
         let view = AddressbookRoleView(model: viewModel)
-        return CoreHostingController(view)
+        return CoreHostingController(view, env: env)
     }
 
 #if DEBUG
--- a/Core/Core/Features/Inbox/ComposeMessage/ComposeMessageAssembly.swift
+++ b/Core/Core/Features/Inbox/ComposeMessage/ComposeMessageAssembly.swift
@@ -23,7 +23,7 @@ import AVKit
 public enum ComposeMessageAssembly {
 
     public static func makeComposeMessageViewController(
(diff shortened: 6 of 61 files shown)


inst-danger:

</details>

inst-danger:

</details>

inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>



#### Release Note: 
Fixed issue of Root account Teacher user unable to access Assignments.

[MBL-18965](https://instructure.atlassian.net/browse/MBL-18965)

<details><summary>:x: XCTest failed: CoreTests/QuizSubmissionBreakdownViewModelTests/testRouting</summary>


    XCTAssertTrue failed
    XCTAssertTrue failed
    XCTAssertTrue failed
</details>

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 4e280224ef4632275cf32cc3867a7ec69974ede1
</p>

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
```diff
@@ -63,6 +63,10 @@ open class AppEnvironment {
         self.logger = Logger.shared
     }
 
+    public var isShared: Bool {
+        api.baseURL == AppEnvironment.shared.api.baseURL
+    }
+
```
vargaat: Using this to decide wether to force refresh kills the cache for those accounts not matching this condition. I'm thinking about using the TTL field to store the instance's baseurl. The `UseCase` reading and writing TTL has the environment so it could prefix the cachekey with the host, separating the cache invalidation time of the two instances.
On the other hand this won't solve the database corruption if the same API call towards different instances write the same coredata entity. The last one writing the DB will win. With the current implementation state the sub-account always wins since it force refreshes all the time. I'm thinking about solutions but without writing the host inside the CoreData entity and filtering to it, I think we can't pull this off.

Summarized:
1. Prefix the `UseCase`'s `cacheKey` with the host of the environment that is saved to `TTL.key`. This will make two separate cache invalidation timers towards the different base urls.
2. Add the host to all CoreData entities that can conflict with each other and modify the use-cases to only return entities that were fetched using the same baseurl. I'm afraid in the end this could affect all CoreData entities so maybe a centralized solution would be preferable.

As a temporary solution we can leave this as-is but we definitely need a better generic solution to this problem. 
Please think about this and share your thoughts.

suhaibabsi-inst: Thanks for sharing that. I think your concerns are totally valid for that regard.

This property is mainly used to overcome the limitation of `Store` in refresh function. Which causes the store not to respond with cached object when this function is used:
https://github.com/instructure/canvas-ios/blob/438aea72231817eeb5d48324c134dca99b6a4a76/Core/Core/Common/CommonModels/Store/UseCase.swift#L80

It was basically used for `GetCourse` and `GetAssignment` use cases. `GetCourse` was scoped properly for the dedicated course using local ID (123) format rather than expanded one (23900000123). So, the entity won't conflict with the one fetched using root base URL. So, no issue can stem from that for the current setup.

The issue here was because `GetCourse` was called before, but at the place where refresh is attempted store wasn't able to use the cached one for that limitation described above, hence trying to force refresh.

A generalized solution for such an anticipated problem can be an extra precaution.
Though it might not be needed for the moment, considering that expanded IDs are used for objects fetched through root API. Which can be used for `UseCase` scoping with no issues.

Let me know your thoughts about those points.

Core/Core/Common/CommonModels/API/ID.swift
```diff
@@ -146,4 +146,15 @@ public extension String {
         formatter.numberStyle = .decimal
         return formatter
     }()
+
+    var asRootID: String {
```
vargaat, line 154: We could add an explanatory comment because just looking at the variable's name it's not clear whether `root` is the root account on server side (when talking about root and child accounts) or root in a way that the user used this domain to log in which could be the child account.d

Core/Core/Features/AttachmentPickers/ViewModel/AudioPickerViewModel.swift
```diff
@@ -56,12 +56,12 @@ class AudioPickerViewModel: NSObject, ObservableObject {
...
 
     private let formatter: DateComponentsFormatter
-    private let router: Router
+    private let env: AppEnvironment
     private var subscriptions = Set<AnyCancellable>()
     private var onSelect: (URL) -> Void
 
-    public init(router: Router, interactor: AudioPickerInteractor, onSelect: @escaping (URL) -> Void = { _ in }) {
-        self.router = router
+    public init(env: AppEnvironment, interactor: AudioPickerInteractor, onSelect: @escaping (URL) -> Void = { _ in }) {
```
vargaat: This seems to be unnecessary, only the router is used that indicates for me that this entity doesn't work with CoreData or the API.

Core/Core/Features/Inbox/InboxSettings/ViewModel/InboxSettingsViewModel.swift
```diff
@@ -36,11 +36,11 @@ public class InboxSettingsViewModel: ObservableObject {
     private let inboxSettingsInteractor: InboxSettingsInteractor
     private var subscriptions = Set<AnyCancellable>()
     private var inboxSettings: CDInboxSettings?
-    private let router: Router
+    private let env: AppEnvironment
 
-    init(interactor: InboxSettingsInteractor, router: Router) {
+    init(interactor: InboxSettingsInteractor, env: AppEnvironment) {
```
vargaat: Only the router is used, seems to be an unnecessary extension to this broader dependency.

Core/Core/Features/Inbox/ViewModel/InboxViewModel.swift
```diff
@@ -178,15 +181,15 @@ public class InboxViewModel: ObservableObject {
...
+            .sink { [router, messageInteractor, weak env] source in
                 // In the parent app we need a different logic for student context picker
                 if messageInteractor.isParentApp {
                     if let bottomSheet = router.match("/conversations/new_message") {
                         router.show(bottomSheet, from: source, options: .modal())
                     }
                 } else {
                     router.show(
-                        ComposeMessageAssembly.makeComposeMessageViewController(),
+                        ComposeMessageAssembly.makeComposeMessageViewController(env: env ?? .shared),
```
vargaat: Could we keep a strong reference to `env`? Falling back to `.shared` seems to be error prone.

Core/Core/Features/Quizzes/QuizDetails/Teacher/ViewModel/TeacherQuizSubmissionBreakdownViewModelLive.swift
```diff
@@ -34,17 +34,17 @@ public class TeacherQuizSubmissionBreakdownViewModelLive: SubmissionBreakdownVie
     private let courseID: String
     private var submissions: Store<GetAllQuizSubmissions>
     private var enrollments: Store<GetEnrollments>
-    private var submissionsPath: String { "courses/\(courseID)/quizzes/\(quizID)/submissions" }
+    private var submissionsPath: String { "/courses/\(courseID)/quizzes/\(quizID)/submissions" }
 
-    init(courseID: String, quizID: String) {
+    init(env: AppEnvironment, courseID: String, quizID: String) {
```
vargaat: `env` could go as the last dependency to its usual place.

Teacher/Teacher/Submissions/SubmissionFilterPickerViewController.swift
```diff
@@ -38,12 +38,14 @@ class SubmissionFilterPickerViewController: UIViewController {
     }
 
     static func create(
+        env: AppEnvironment,
```
vargaat: Can we move this to after `filter`?

Core/Core/Features/Assignments/AssignmentDetails/AssignmentSubmissionBreakdownViewModel.swift
```diff
@@ -41,15 +41,15 @@ public class AssignmentSubmissionBreakdownViewModel: SubmissionBreakdownViewMode
     private let courseID: String
     private let submissionTypes: [SubmissionType]
     private var summary: Store<GetSubmissionSummary>
-    private var submissionsPath: String { "courses/\(courseID)/assignments/\(assignmentID)/submissions" }
+    private var submissionsPath: String { "/courses/\(courseID)/assignments/\(assignmentID)/submissions" }
 
-    public init(courseID: String, assignmentID: String, submissionTypes: [SubmissionType], color: UIColor? = nil) {
+    public init(env: AppEnvironment, courseID: String, assignmentID: String, submissionTypes: [SubmissionType], color: UIColor? = nil) {
```
rh12: Please move env to the end

Core/Core/Features/Assignments/AssignmentEditor/AssignmentAssigneeList.swift
```diff
@@ -42,10 +42,10 @@ struct AssigmentAssigneeList: View {
         !isEveryoneMatching && sections.isEmpty && groups.isEmpty && students.isEmpty
     }
 
-    init(courseID: String, groupCategoryID: String?, selection: Binding<[Assignee]>) {
-        groups = AppEnvironment.shared.subscribe(GetGroupsInCategory(groupCategoryID))
-        sections = AppEnvironment.shared.subscribe(GetCourseSections(courseID: courseID))
-        students = AppEnvironment.shared.subscribe(GetContextUsers(context: .course(courseID), type: .student))
+    init(env: AppEnvironment, courseID: String, groupCategoryID: String?, selection: Binding<[Assignee]>) {
```
rh12: Please move env to the end

Core/Core/Features/Assignments/AssignmentEditor/AssignmentAssigneePicker.swift
```diff
@@ -40,16 +40,16 @@ struct AssignmentAssigneePicker: View {
         groups.state == .loading || sections.state == .loading || students.state == .loading
     }
 
-    init(courseID: String, groupCategoryID: String?, overrides: Binding<[Override]>, override: Override) {
+    init(env: AppEnvironment, courseID: String, groupCategoryID: String?, overrides: Binding<[Override]>, override: Override) {
```
rh12: Please move env to the end

Core/Core/Features/Inbox/Addressbook/AddressbookAssembly.swift
```diff
@@ -23,7 +23,7 @@ import UIKit
 public enum AddressBookAssembly {
 
     public static func makeAddressbookRecipientViewController(
-        env: AppEnvironment = .shared,
+        env: AppEnvironment,
```
rh12: Please move env to the end

Core/Core/Features/Inbox/Addressbook/AddressbookAssembly.swift
```diff
@@ -40,11 +40,11 @@ public enum AddressBookAssembly {
             selectedRecipients: selectedRecipients
         )
         let view = AddressbookRecipientView(model: viewModel)
-        return CoreHostingController(view)
+        return CoreHostingController(view, env: env)
     }
 
     public static func makeAddressbookRoleViewController(
-        env: AppEnvironment = .shared,
+        env: AppEnvironment,
```
rh12: Please move env to the end

Core/Core/Features/Inbox/ComposeMessage/ComposeMessageAssembly.swift
```diff
@@ -23,7 +23,7 @@ import AVKit
 public enum ComposeMessageAssembly {
 
     public static func makeComposeMessageViewController(
-        env: AppEnvironment = .shared,
+        env: AppEnvironment,
```
rh12: Please move env to the end

Core/Core/Features/Inbox/ComposeMessage/ViewModel/ComposeMessageViewModel.swift
```diff
@@ -109,15 +109,15 @@ final class ComposeMessageViewModel: ObservableObject {
 
     // MARK: Public interface
     public init(
-        router: Router,
+        env: AppEnvironment,
```
rh12: Please move env to the end

Core/Core/Features/Inbox/MessageDetails/ViewModel/MessageDetailsViewModel.swift
```diff
@@ -50,18 +50,18 @@ class MessageDetailsViewModel: ObservableObject {
     // MARK: - Private
     private var subscriptions = Set<AnyCancellable>()
     private let interactor: MessageDetailsInteractor
-    private let router: Router
+    private let env: AppEnvironment
     private let myID: String
     private let allowArchive: Bool
 
-    public init(router: Router, interactor: MessageDetailsInteractor, myID: String, allowArchive: Bool) {
+    public init(env: AppEnvironment, interactor: MessageDetailsInteractor, myID: String, allowArchive: Bool) {
```
rh12: Please move env to the end

rh12: It seems to me `router` would be enough

Parent/Parent/Assignments/View/AssignmentDetailsViewController.swift
```diff
@@ -68,13 +68,15 @@ class AssignmentDetailsViewController: UIViewController, CoreWebViewLinkDelegate
     }
 
     static func create(
+        env: AppEnvironment,
```
rh12: Please move env to the end

Core/Core/Features/Inbox/MessageDetails/ViewModel/MessageDetailsViewModel.swift
suhaibabsi-inst: We need this for line 191:
```swift
                ComposeMessageAssembly.makeComposeMessageViewController(
                    env: env,
```

rh12 (approved):
QA + 1 (issue in ticket is fixed, smoked Teacher, smoked Student)

rh12 (approved):

PR #3473, opened 2025-07-02
Assignment List Preferences - Grading Period filter behaviour change

ndrsszsz:
## Assignment List Preferences - Grading Period filter behaviour change

refs: MBL-18974
affects: Teacher, Student
release note: None
test plan: See ticket.

If there is only 1 active grading period, we are showing the currently active grading period by default (which is correct), but if there's only 1 grading period for the course then we are not showing the grading period filter options on the preferences screen. This can cause the following problem: If there is a currently active grading period and the user has an assignment outside of that, then it will not be displayed on the assignment list.

Solution requested by product:
Show the assignment filter options even when there is only 1 grading period.

## Test

Account and user for testing the above mentioned scenario is available in the ticket.

## Checklist

- [ ] Tested


inst-danger:

</details>

inst-danger:

</details>

PR #3484, opened 2025-07-08
Teacher - Teachers of large courses are unable to access all submissions directly from the To Do list

ndrsszsz:
## Teacher - Teachers of large courses are unable to access all submissions directly from the To Do list

refs: MBL-18971
affects: Teacher
release note: None
test plan: See ticket.

## Test Plan

So the Submission List screen that can be accessed from the Assignment Details screen uses it's own separate interactor, and so does Speedgrader (that - in this case - gets opened from the todo list by tapping on the assignment).
SubmissionListInteractor has ignoreCache parameter set to true, so I added this to the Speedgrader one too which solved the issue. I'm not pleased with this fix but unfortunately I couldn't come up with a better one right now.

As I see, the root cause of issues like this is that we handle/transform/filter submission data in more than one places in the code, each can have different ways of handling data, and this can cause bugs like this one. The best would be to have only one interactor for submissions and that one should be user everywhere where it's needed throughout the app.

1. Open the app with fresh start (without cached data).
2. Log in with the user in the ticket.
3. Navigate directly to the todo tab without going anywhere else in the app.
4. Tap "assignment 1".
5. Speedgrader should be opened.
6. If you swipe between the students long enough (an eternity) then you can see Ruth Garner's submission.

Because of the eternity thing mentioned in the 6th step, I tested this in the following way:
Check out the code and add a filter for the submissions that filters out Ruth Garner's one and only sends that one forward. So when you tap on the assignment on the Todo tab and Speedgrader gets opened, you will see the submission right away without swiping through a lot of students.

Additional info: In case of this much number of student submissions, Speedgrader is annoyingly slow and sometimes it even freezes for like 10 seconds when trying to swipe to the next student.

## Checklist

- [ ] Tested

### DIFF
--- a/Teacher/Teacher/SpeedGrader/StudentPager/Model/SpeedGraderInteractorLive.swift
+++ b/Teacher/Teacher/SpeedGrader/StudentPager/Model/SpeedGraderInteractorLive.swift
@@ -161,7 +161,7 @@ class SpeedGraderInteractorLive: SpeedGraderInteractor {
         let submissionsUseCase = GetSubmissions(context: context, assignmentID: assignmentID, filter: filter)
         submissionsUseCase.shuffled = anonymizeStudents
         return ReactiveStore(useCase: submissionsUseCase, environment: env)
-            .getEntities(loadAllPages: true)
+            .getEntities(ignoreCache: true, loadAllPages: true)
             .eraseToAnyPublisher()
     }
 


vargaat (approved):
QA+1, I was able to swipe to Ruth's submission and seen al the pages getting loaded. Actually, I agree that we should force refresh here, it's important for teachers to see the latest state of student submission. The initial loading is pretty slow because each page takes more than 6 seconds to load and there are 3 full pages. We could improve this in the future by not loading all submission details only when the teacher swipes to it.

rh12 (approved):

rh12 (approved):

PR #3508, opened 2025-07-18
SpeedGrader - Details tab

rh12:
refs: [MBL-18907](https://instructure.atlassian.net/browse/MBL-18907), [MBL-19054](https://instructure.atlassian.net/browse/MBL-19054)
affects: Teacher
release note: Added Student Notes and Word Count to SpeedGrader. Improved SpeedGrader drawer opening/closing behavior.

## What's new
- Added Details tab
  - Added Word Count
  - Added Student Notes
  - Handled empty state
- Fixed drawer button functionality (see [design](https://www.figma.com/design/d0XYwwGe37hEWdmArVVBhr/%E2%8F%B3-Mobile-Design-2024-2025?node-id=9921-51336&t=BIwlTr9qcci2gGWu-4) for details)

### Chores
- Moved SpeedGraderPage sub-viewmodel creation to assembly
- Added comments & marks to `UseCase`
- Added `InstUI.CollapsibleListSection`
  - it uses the new `InstUI.SectionDisclosureStyle`
- Added `InstUI.Divider(isLast:)`
- Added `InstUI.TopDivider`
- Updated `InstUI.Styles.Elevation`
- Added CoreTester to Teacher & Parent build targets (to fix failing SwiftUI previews)

## What's not changed
- Grading History will be done in separate task

## Test plan
- Verify WordCount, Student Notes, Empty states according to ticket.
- Verify drawer buttons work as described in Figma (linked above)

## Screenshots
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist
- [ ] A11y checked
- [ ] Tested on phone
- [x] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-18907]: https://instructure.atlassian.net/browse/MBL-18907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MBL-19054]: https://instructure.atlassian.net/browse/MBL-19054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


inst-danger:

</details>

Core/Core/Common/CommonUI/InstUI/Views/Divider.swift
```diff
@@ -45,7 +49,22 @@ extension InstUI {
...
+
+        private let style: Divider.Style
+
+        public init(_ style: Divider.Style = .full) {
+            self.style = style
+        }
+
+        public var body: some View {
+            InstUI.Divider(style)
+                .offset(y: -1)
```
vargaat, line 70: Can we add a comment on why is this required? I would assume that if this divider is the first one in a VStack and it's clipped then this divider would draw outside of the view and become invisible.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,121 @@
...
+    private let courseId: String
+
+    init(courseId: String) {
+        self.courseId = courseId
+    }
+
+    // MARK: - Custom Columns
+
+    /// Loads all Custom Columns data from API into CoreData, ignoring cache.
+    /// It loads entries from all columns, for all students.
```
vargaat, line 52: I would move this comment to the protocol since that's what we'll have in places where we use this interactor.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,121 @@
...
+import Combine
+import Core
+import Foundation
+
+protocol CustomGradebookColumnsInteractor {
+    func loadCustomColumnsData() -> AnyPublisher<Void, Error>
+    func getStudentNotesEntries(userId: String) -> AnyPublisher<[StudentNotesEntry], Error>
+}
+
+final class CustomGradebookColumnsInteractorLive: CustomGradebookColumnsInteractor {
```
vargaat, line 39: There are lot of methods in this class not available in the protocol. Those can be private I guess and separated from the public interface. (Maybe static as well because those methods stateless and we could get rid of the weak self but we would need to pass `courseId` along so it depends...)

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,121 @@
...
+//
+
+import Combine
+import Core
+import Foundation
+
+protocol CustomGradebookColumnsInteractor {
+    func loadCustomColumnsData() -> AnyPublisher<Void, Error>
+    func getStudentNotesEntries(userId: String) -> AnyPublisher<[StudentNotesEntry], Error>
+}
```
vargaat: I often end up interactor protocols and live implementations like this. There's a `courseId` parameter in the live implementation that's needed for `loadCustomColumnsData` but that's missing from the protocol and without this reading only the protocol alone isn't enough to understand what it's exactly doing. (Fetching custom columns data for what?) Adding the init to the protocol might be counter intuitive as it would dictate how the implementation should look so I just ended up adding a `get` only variable to flag that there's a must have parameter for the job.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,121 @@
...
+    init(courseId: String) {
+        self.courseId = courseId
+    }
+
+    // MARK: - Custom Columns
+
+    /// Loads all Custom Columns data from API into CoreData, ignoring cache.
+    /// It loads entries from all columns, for all students.
+    func loadCustomColumnsData() -> AnyPublisher<Void, Error> {
+        return getCustomColumns(ignoreCache: true)
```
vargaat, line 56: Why the hardcoded ignore cache `true`? Could we control this from outside so it would be explicit for readers that we're doing a force refresh?

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,121 @@
...
+            .flatMap { [weak self] columns -> AnyPublisher<Void, Error> in
+                guard let self else { return Publishers.typedEmpty() }
```
vargaat: A code piece like the one below would never finish despite the subscription being stored, because the interactor deallocates and the empty() publisher never finishes the stream. I think in such cases the returned cancellable should control the cancellation of the task and if it needs the interactor to finish it, then it should strongly retain it.
```
CustomGradebookColumnsInteractorLive()
    .loadCustomColumnsData()
    ...
    .store(in: &subscriptions)
```

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,121 @@
...
+            .map { columns in
+                columns.filter { $0.isTeacherNotes && !$0.isHidden }
+            }
+            .eraseToAnyPublisher()
+    }
+
+    func getStudentNotesEntries(userId: String) -> AnyPublisher<[StudentNotesEntry], Error> {
+        getStudentNotesColumns()
+            .flatMap { [weak self] columns -> AnyPublisher<[StudentNotesEntry], Error> in
+                guard let self else { return Publishers.typedEmpty() }
```
vargaat: The same weak self thingy here as above.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/ViewModel/StudentNotesViewModel.swift
```diff
@@ -0,0 +1,53 @@
...
+    // MARK: - Init
+
+    init(
+        userId: String,
+        interactor: CustomGradebookColumnsInteractor
+    ) {
+        self.interactor = interactor
+
+        interactor.getStudentNotesEntries(userId: userId)
+            .receive(on: RunLoop.main)
```
vargaat: CombineSchedulers could be used for testing purposes.

Teacher/Teacher/SpeedGrader/Details/SubmissionWordCount/Model/SubmissionWordCountInteractor.swift
```diff
@@ -0,0 +1,57 @@
...
+import Core
+import Foundation
+
+protocol SubmissionWordCountInteractor {
+    func getWordCount(userId: String, attempt: Int) -> AnyPublisher<Int?, Error>
+}
+
+final class SubmissionWordCountInteractorLive: SubmissionWordCountInteractor {
+
+    private let assignmentId: String
```
vargaat: Consider adding the `assignmentId` to the protocol as a variable for the same reasons as at the other interactor.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
rh12: I prefer them near the actual logic, but it also make sense to have them at the protocol definition. Ended up duplicating them.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
rh12: I made them private for now, with one exception which I moved to the protocol.
(May make them internal if it makes testing simpler)
I kept the logical order, regardless of access level.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
rh12: As discussed I will follow this pattern

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
rh12: As discussed it's inteded to be called once per SpeedGrader session.
Added comment to elaborate.
This may change later when we optimize requests.

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
rh12: The idea was that since these interactors are stored for the whole SG session, they wouldn't be released.
But you're right, it's not that robust this way.
Ended up capturing self explicitly with `[self]` rather then omitting it to indicate this is on purpose

Teacher/Teacher/SpeedGrader/Details/CustomGradebookColumns/Model/CustomGradebookColumnsInteractor.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    /// Fetches all Student Notes entries for the given `userId`.
+    func getStudentNotesEntries(userId: String) -> AnyPublisher<[StudentNotesEntry], Error>
+}
+
+final class CustomGradebookColumnsInteractorLive: CustomGradebookColumnsInteractor {
+
+    let courseId: String
+
+    init(courseId: String) {
```
suhaibabsi-inst: Let's pass env to this, and use it with ReactiveStore s

balintbartok (dismissed):
QA + 1

PR #3514, opened 2025-07-18
[Horizon] WrappingHStack Improvement

reabbotted:
This is not perfect, but it is an improvement. It keeps items from overlapping each other, but doesn't allow for vertically centering row items.

Simulator Screenshot - iPhone 16 Pro Max - 2025-07-18 at 13 37 53
Simulator Screenshot - iPhone 16 Pro Max - 2025-07-18 at 14 35 40

[ignore-commit-lint]
affects: Horizon


Ahmed-Naguib93 (dismissed):
Works fine, and also tested on the Student app.

PR #3515, opened 2025-07-20
User logged in QED Root account can't access files when opened from Modules

suhaibabsi-inst:
refs: MBL-18966
affects: Student, Teacher
release note: none.

## Analysis

After thorough investigation and multiple failed attempts to try reproduce the issue, by searching the code about what might lead to the exact experience shown on the ticket. It was concluded that the only execution path that could lead to that experience (popping back, or showing screen with logo on iPad upon load finish) must be go through this line:

https://github.com/instructure/canvas-ios/blob/e02cca4ec86bbffa64dfb4169acb475394b39c85/Core/Core/Features/Files/View/FileDetails/FileDetailsViewController.swift#L206

Which, as the condition suggesting, can happen for the following scenarios:
1. API fetch completed with valid response but context reports saved objects to `Store.frc: NSFetchedResultsController` after fetch callback is being called. 

https://github.com/instructure/canvas-ios/blob/e02cca4ec86bbffa64dfb4169acb475394b39c85/Core/Core/Common/CommonModels/Store/Store.swift#L317

2. API fetch completed successfully with empty response (no files), no error is reported.
3. Offline check yield `true`, for possible instability of network connection at the time of requesting Store data:

https://github.com/instructure/canvas-ios/blob/e02cca4ec86bbffa64dfb4169acb475394b39c85/Core/Core/Common/CommonModels/Store/Store.swift#L308

## Resolution

Even with using Apple's tool `Network Link Conditioner` to simulate poor internet connectivity, I haven't been able to reproduce the issue. Thus, my approach to this would be to put it up few precautions for possible issues:

1. Move execution of API fetch callback to be always on Main thread.
2. Have some logic to defend against the case where fetched objects being reported after API fetch callback get finished.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Files/View/FileDetails/FileDetailsViewController.swift
+++ b/Core/Core/Features/Files/View/FileDetails/FileDetailsViewController.swift
@@ -220,8 +220,8 @@ public class FileDetailsViewController: ScreenViewTrackableViewController, CoreW
                 RemoteLogger
                     .shared
                     .logError(
-                        name: "Popping back File details view: \(fileRoute)",
-                        reason: "Unknown state upon loading file information"
+                        name: "Failed to show File details view. Popping back to previous screen.",
+                        reason: "Unknown state upon loading (\(fileRoute)) file information"
                     )
 
                 // File was deleted, go back.


inst-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please add a release note. If no release note is wanted, use `none`. Example: `release note: Fixed a bug that prevented users from enjoying the app.`
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Please add which apps this change affects. Example: `affects: Teacher, Student` or `affects: none`
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Please add a reference to a Jira ticket. For example: `refs: MBL-10023`
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


**Canvas iOS** | 91.37% | 91.36% | 0.01%
Core/Core/Common/CommonUI/CoreWebView/View/CoreWebViewAccessibilityHelper.swift | 0% | 0% | 0%
Student/Student/WidgetRouting/WidgetRouter+GradeListWidget.swift | 0% | 0% | 0%
<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against db8a96f8391bc224474d02a827e4fc38ec537765
</p>

inst-danger:

</details>

vargaat (dismissed):
I also take a look on this bug but have no better idea on how to solve this because I also wasn't able to reproduce it and because the FileDetailsViewController is such a mess that it's hard to follow up what's happening. I was eyeing with moving the fetch's completion to the main thread in the past as well so I think this is a nice improvement. I'm not sure about if the new helper bool will make things better.

1. One way to improve the situation would be to rewrite FileDetailsViewController or at least the file fetch using ReactiveStore but this could be time consuming.
2. We could also log breadcrumbs at important states and just log an error when we reach the point where the view controller is popped. The error would list the breadcrumbs so we could get the state of the logic. (RemoteLogger)
3. We could improve the experience by proper error checks for deleted files and if we receive empty API responses or we are at some unclear state we could show an error alert with a retry dialog. If this is really related to network connectivity issues a retry might solve this.

Core/Core/Features/Files/View/FileDetails/FileDetailsViewController.swift
```diff
@@ -199,20 +204,34 @@ public class FileDetailsViewController: ScreenViewTrackableViewController, CoreW
...
-            } else if files.requested, !files.pending, localURL == nil {
+            } else if files.requested,
+                      !files.pending,
+                      !files.isToBeUpdated,
+                      localURL == nil {
+
+                RemoteLogger
+                    .shared
+                    .logError(
+                        name: "Popping back File details view: \(fileRoute)",
```
rh12: 1. We usually move the dynamic part to the `reason` to allow for Crashlytics to group these events together. Otherwise we would get a different error event for each file.
2. Could we add some context to this error name? Like "Failed to show File details view. Popping back to previous screen.

PR #3538, opened 2025-07-30
[Horizon] Display icons alongside highlights

Ahmed-Naguib93:
https://instructure.atlassian.net/browse/CLX-2255

affects: Horizon

### DIFF
--- a/Horizon/Horizon/Resources/WebHighlighting.js
+++ b/Horizon/Horizon/Resources/WebHighlighting.js
@@ -13,153 +13,880 @@
 // GNU Affero General Public License for more details.
 //
 // You should have received a copy of the GNU Affero General Public License
-// along with this program.  If not, see <https://www.gnu.org/licenses/>.
-//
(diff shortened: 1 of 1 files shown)


inst-danger:

</details>

reabbotted:
We don't really want to just have the WebHighlighting.js file edited manually into JavaScript. The way this part of the project is structured is a Typescript project so that we get the benefits of TypeScript. We then compile the TypeScript project into minified JavaScript. Take a look at the README.md. 

I had mostly made my way through this feature too. You can see my branch here:
https://github.com/instructure/canvas-ios/tree/CLX-2255-icons-should-be-displayed-alongside-highlights-for-accessibility-purposes

Something we don't want to happen is if the icons are updated on the iOS side for them to get out of sync with the web side, so I deliver the SVG content across from iOS to web. Unfortunately, I wasn't able to just read the SVG file from the iOS Assets bundle so instead what I've done is placed the configuration in the same place so that it's clear in the future that if one changes, the other must also change.

PR #3543, opened 2025-07-31
SpeedGrader - Add support for custom grading schemes

rh12:
refs: [MBL-19074](https://instructure.atlassian.net/browse/MBL-19074)
affects: Teacher, Student, Parent
release note: none

See ticket for details

Out of scope: GradePicker respecting grading scheme selection in assignment itself (it has it's own task)

(tests will follow in the followup for the main grade input task)

## TestPlan
- Set a custom grading scheme in course details
- Do NOT set a different one in the assignment (leave it on default)
- Verify these kinds of grading schemes display and work properly for all kinds of Letter Grade and GPA assignments:
  - Percentages based which has decimal values
  - Point based matching the assignment's max score
  - Point based not matching that
  - Point based which also have decimals
- Compare GradePicker subtitles with grading scheme displays in Manage Garding Scheme page

- Smoke test total grade in Student app with percent & point based schemes

## Checklist
- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-19074]: https://instructure.atlassian.net/browse/MBL-19074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Features/Grades/Model/Entities/GradingSchemes/PercentageBasedGradingScheme.swift
+++ b/Core/Core/Features/Grades/Model/Entities/GradingSchemes/PercentageBasedGradingScheme.swift
@@ -20,21 +20,26 @@ import Foundation
 
 public struct PercentageBasedGradingScheme: GradingScheme {
 
+    /// converts 0.1234 -> "12.34%", 0.42 -> "42%"
+    private static let percentFormatter = GradeFormatter.percentFormatter
+
     public let entries: [GradingSchemeEntry]
+
+    public init(entries: [GradingSchemeEntry]) {
+        self.entries = entries
+    }
+
+    public var maxFormattedValue: String? {
+        Self.percentFormatter.string(from: NSNumber(value: 1))
+    }
+
+    public func formattedEntryValue(_ value: Double) -> String? {
+        Self.percentFormatter.string(from: NSNumber(value: value))
+    }
 
     public func formattedScore(from value: Double) -> String? {
         Self.percentFormatter.string(from: NSNumber(value: value))
     }
-
-    private static let percentFormatter: NumberFormatter = {
-        let formatter = NumberFormatter()
-        formatter.numberStyle = .percent
-        formatter.decimalSeparator = "."
-        formatter.multiplier = 1
-        formatter.maximumFractionDigits = 3
-        formatter.roundingMode = .down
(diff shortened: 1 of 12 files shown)


inst-danger:

</details>

ndrsszsz (approved):
QA +1
Looks good to me.

Core/Core/Features/Grades/Model/Entities/GradingSchemes/PercentageBasedGradingScheme.swift
```diff
@@ -20,21 +20,29 @@ import Foundation
...
     public let entries: [GradingSchemeEntry]
 
-    public func formattedScore(from value: Double) -> String? {
-        Self.percentFormatter.string(from: NSNumber(value: value))
+    public init(entries: [GradingSchemeEntry]) {
+        self.entries = entries
+    }
+
+    public var formattedMaxValue: String? {
+        Self.percentFormatter.string(from: NSNumber(value: 1))
```
suhaibabsi-inst, line 32: No need to change that for now.

But I was thinking, perhaps, utilizing the new `FormatStyle` type would help us a lot eliminate the need to persist those formatter objects.

For example:
https://developer.apple.com/documentation/foundation/formatstyle/percent-6cwuv

suhaibabsi-inst (commented):
Looks good to me,

One issue though, it seems working when testing an assignment using a Point-based grading scheme defined on the account level. While if you use a point-based grading scheme defined at the point of writing the assignment, points get translated correctly to letter grade, but for GradePicker, range is till being shown in percentage.

suhaibabsi-inst (approved):

Core/Core/Features/Grades/Model/Entities/GradingSchemes/PercentageBasedGradingScheme.swift
rh12: Thanks, that may be worth looking into!
We should make sure to define the few format styles we use for grades somewhere though.

PR #3550, opened 2025-08-05
Horizon App Merge

szabinst:
Merges back changes from `feature/horizon` with the addition of a unified login experience. This includes the whole Horizon app, but I'd suggest to ignore the Horizon and HorizonUI folder and focus on reviewing changes in `Core`, `Student` and everything else. 

## What's new
- Refactored StudentAppDelegate so async calls can be chained 
- Added error handling to the login flow so now the loading screen can display an alert 
- Added a new call to decide which [App Experience](#app-experience) to load. 
- Horizon is a `framework` instead of an `application` and now it's referenced from within the Student project the same way as `Core` is referenced

<a name="app-experience"></a>
## App Experience
- The Student app now has two experiences called `Academic` and `Career`.
- The current backend implementation is tailored for web and it's based on sessions. 
- This is not suited for native apps, because they regularly renew sessions.
- In order to overcome this issue, we incorporate `UserDefaults` for now in the following way: 
  - If the user is logging in for the first time, we query the `ExperienceSummary` API to retrieve the `currentApp` and `availableApps`.
  - We save the `currentApp` to `UserDefaults`
  - If it's a regular login (like an app start) we get the saved preference from `UserDefaults`. 
  - To enable the App Experience Switch, we rely on the `availableApps` list that is retrieved from Core Data 

## Screenshots

<table>
<tr><th>Academic - Profile Settings View</th><th>Career - Account View</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


inst-danger:

</details>

inst-danger:

</details>

szabinst:
@balintbartok can you tell me a bit more? did you log in with an account that has only academic available app?

Core/Core/Features/Profile/Settings/ProfileSettingsViewController.swift
```diff
@@ -114,6 +122,49 @@ public class ProfileSettingsViewController: ScreenViewTrackableViewController {
...
+        NSLayoutConstraint.activate([
+            indeterminateLoadingIndicator.centerXAnchor.constraint(equalTo: loadingIndicatorContainerView.centerXAnchor),
+            indeterminateLoadingIndicator.centerYAnchor.constraint(equalTo: loadingIndicatorContainerView.centerYAnchor),
+            indeterminateLoadingIndicator.widthAnchor.constraint(equalToConstant: 40),
+            indeterminateLoadingIndicator.heightAnchor.constraint(equalToConstant: 40)
+        ])
+    }
+
+    private func animateLoadingIndicator(show: Bool) {
+        weak var weakSelf = self
```
suhaibabsi-inst, line 141: This is interesting use of weak self!

Core/Core/Features/Profile/Settings/ProfileSettingsViewController.swift
```diff
@@ -122,57 +173,63 @@ public class ProfileSettingsViewController: ScreenViewTrackableViewController {
     }
 
     func reloadData() {
-        var channelTypes: [CommunicationChannelType: [CommunicationChannel]] = [:]
-        for channel in channels {
-            channelTypes[channel.type] = channelTypes[channel.type] ?? []
-            channelTypes[channel.type]?.append(channel)
-        }
+        Task {
```
suhaibabsi-inst: Perhaps you need to change that to be `Task { @MainActor ... }` so you guarantee is being executed on the main thread. That's because UI code is being called in this task. Mainly `tableView.reloadData()` and `tableView.refreshControl?.endRefreshing()` functions.

Student/Student/StudentAppDelegate.swift
```diff
@@ -283,30 +263,222 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
     }
 }
 
+// MARK: - Setup User Environment
```
suhaibabsi-inst, line 266: How about moving this extension to a separate file. Also, perhaps we can make attempt to lift some of these out of the delegate object if not accessing any property on it, so to avoid keep referencing self in `ReactiveStore(useCase: GetUserProfile())` publisher operators stack.

suhaibabsi-inst (commented):
Thanks @szabinst !
In General, looks good to me.
Just a couple of notes for your consideration.

Student/Student/StudentAppDelegate.swift
szabinst: The Horizon app did that lift and moved these responsibilities to a new class called `SessionInteractor`, but I didn't want to introduce such a big change with this PR. In the long term, I'd definitely prefer to strip down the AppDelegates but that's a task on its' own.

Core/Core/Features/Profile/Settings/ProfileSettingsViewController.swift
szabinst: If you declare a weak or unowned reference at the function's local scope, you don't have to recapture that reference in multiple and/or nested closures so the code stays a bit more readable.

Core/Core/Features/Profile/Settings/ProfileSettingsViewController.swift
szabinst: Good point! Pushed the change.

suhaibabsi-inst (commented):
Code +1

Student/Student/Localizable.xcstrings
rh12, line 1: I think this will basically revert the [latest translation PR](https://github.com/instructure/canvas-ios/pull/3548/files#diff-3a908e390f62573a10b5ac6cb12be7446d77c3a9b1e1c159f0720fb6a24ad07b). Could you please resolve this?

Student/Student/StudentAppDelegate.swift
```diff
@@ -46,11 +54,14 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
     private lazy var gradeListWidgetRouter = WidgetRouter.createGradeListRouter()
     private lazy var courseGradeWidgetRouter = WidgetRouter.createCourseGradeRouter()
 
-    private lazy var analyticsTracker: PendoAnalyticsTracker = {
+     private lazy var analyticsTracker: PendoAnalyticsTracker = {
```
rh12: Please revert that space

Student/Student/StudentAppDelegate.swift
```diff
@@ -283,30 +263,222 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
...
+                }
+                return Empty().setFailureType(to: Error.self).eraseToAnyPublisher()
+            } else {
+                return Just(()).setFailureType(to: Error.self).eraseToAnyPublisher()
+            }
+        }
+        .eraseToAnyPublisher()
+    }
+
+    func listenForExperienceChanges() {
```
rh12:
```suggestion
    private func listenForExperienceChanges() {
```

Student/Student/StudentAppDelegate.swift
```diff
@@ -283,30 +263,222 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
...
+        AppEnvironment.shared.experience
+            .dropFirst()
+            .sink { [weak self] in
+                self?.setTabBarControllerFor(experience: $0, isStartup: false)
+            }
+            .store(in: &subscriptions)
+    }
+
+    func setTabBarControllerFor(experience: Experience, isStartup: Bool) {
```
rh12:
```suggestion
    private func setTabBarControllerFor(experience: Experience, isStartup: Bool) {
```

Student/Student/StudentAppDelegate.swift
```diff
@@ -283,30 +263,222 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
...
+extension StudentAppDelegate {
+    private func setupUserEnvironment() -> AnyPublisher<Void, Error> {
+        PageViewEventController.instance.userDidChange()
+        updateInterfaceStyle(for: window)
+        CoreWebView.keepCookieAlive(for: environment)
+        PushNotificationsInteractor.shared.userDidLogin(api: environment.api)
+
+        return Just(())
+            .setFailureType(to: Error.self)
+            .eraseToAnyPublisher()
```
rh12:
```suggestion
        return Publishers.typedJust()
```
No need to change if you prefer the original, but feel free to check those helpers out.
Same for other places.

Student/Student/StudentAppDelegate.swift
```diff
@@ -283,30 +263,222 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
...
+                isLoginTransition: false,
+                viewController: LoginNavigationController.create(
+                    loginDelegate: self,
+                    app: .student
+                )
+            )
+        } else {
+            let alert = UIAlertController(
+                title: "Oops, something went wrong",
+                message: "There was an error while logging you in. You can try again, or come back a bit later.",
```
rh12: Shouldn't we localize these two?

Core/Core/Resources/Assets.xcassets/icons/swap_horiz.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
...
+      "filename" : "swap_horiz.svg",
+      "idiom" : "universal"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "properties" : {
+    "template-rendering-intent" : "template"
```
rh12, line 14: I think it should have `"preserves-vector-representation" : true,` like the others

Core/Core/Common/Extensions/CoreData/StringSetTransformer.swift
```diff
@@ -0,0 +1,57 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Foundation
+
+@objc(StringSetValueTransformer)
+final class StringSetValueTransformer: ValueTransformer {
```
rh12: Please update the filename to match the classname

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
```diff
@@ -38,6 +39,7 @@ open class AppEnvironment {
     )
 
     public var app: App?
+    public var experience = CurrentValueRelay<Experience>(.academic)
```
rh12, line 42: Just to doublecheck, wasn't there some memory issue with relays?
My understanding was that we should use subjects instead, but I may be wrong.

Core/Core/Common/Extensions/CoreData/StringSetTransformer.swift
```diff
@@ -0,0 +1,57 @@
...
+        archiver.encode(array, forKey: NSKeyedArchiveRootObjectKey)
+        return archiver.encodedData
+    }
+
+    // Transforms NSData back to Set<String>
+    override func reverseTransformedValue(_ value: Any?) -> Any? {
+        guard let data = value as? Data else { return nil }
+
+        // swiftlint:disable:next force_try
+        let unarchiver = try! NSKeyedUnarchiver(forReadingFrom: data)
```
rh12: Just doublechecking: Is this force unwrapped because `requiringSecureCoding` ensures it will always succeed?
Still, wouldn't it be safer to just guard it and return nil (or assert/log)?

Core/Core/Features/Experience/Data/API/PostSwitchExperienceRequest.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public let method: APIMethod = .post
+    public var path: String {
+        "career/switch_experience"
+    }
+
+    public var body: Body?
+
+    init(experience: String) {
```
rh12:
```suggestion
    public init(experience: String) {
```

Core/Core/Features/Experience/Data/API/APIExperienceSummary.swift
```diff
@@ -0,0 +1,37 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+public struct APIExperienceSummary: Codable, Equatable {
+    let current_app: Experience
+    let available_apps: [Experience]
+}
+
+public enum Experience: String, Codable, Equatable {
```
rh12: Could we move this to it's own file? It's both an APIModel and a regular model, but I guess that's fine, I would prefer simply to not "hide" it inside here.

Core/Core/Features/Experience/Data/GetExperienceSummaryUseCase.swift
```diff
@@ -0,0 +1,33 @@
...
+import CoreData
+import Foundation
+
+public struct GetExperienceSummaryUseCase: APIUseCase {
+    public typealias Model = CDExperienceSummary
+    public let cacheKey: String? = "experience-summary"
+    public let request = GetExperienceSummaryRequest()
+
+    public func write(response: APIExperienceSummary?, urlResponse _: URLResponse?, to client: NSManagedObjectContext) {
+        guard let response = response else { return }
```
rh12:
```suggestion
        guard let response else { return }
```

Core/Core/Features/Experience/Data/SwitchExperienceUseCase.swift
```diff
@@ -0,0 +1,40 @@
...
+        self.cacheKey = nil
+    }
+
+    public func write(
+        response: PostSwitchExperienceRequest.Body?,
+        urlResponse _: URLResponse?,
+        to client: NSManagedObjectContext
+    ) {
+        guard let response else { return }
+        CDExperienceSummary.update(experience: response.experience, in: client)
```
rh12, line 38: is it possible that we don't have the CD entity in CoreData at this point?
Shouldn't we fallback to `save()` it in that case? Or just `save()` it all the time to make sure?

Core/Core/Features/Experience/Domain/ExperienceSummaryInteractor.swift
```diff
@@ -0,0 +1,93 @@
...
+    func isExperienceSwitchAvailableAsync() async -> Bool
+    func switchExperience(to experience: Experience) -> AnyPublisher<Void, Never>
+}
+
+public final class ExperienceSummaryInteractorLive: ExperienceSummaryInteractor {
+    private let environment: AppEnvironment
+    private let scheduler: AnySchedulerOf<DispatchQueue>
+
+    public init(
+        environment: AppEnvironment = AppEnvironment.shared,
```
rh12, line 35: I'm not sure, but shouldn't we always require the environment, because of Consortia issues? (summoning @suhaibabsi-inst )
Call sites could simply pass in `.shared` as a fallback, but it would be required to think about it a bit at least.

Core/Core/Features/Experience/Domain/ExperienceSummaryInteractor.swift
```diff
@@ -0,0 +1,93 @@
...
+            }
+            .eraseToAnyPublisher()
+    }
+
+    public func isExperienceSwitchAvailable() -> AnyPublisher<Bool, Never> {
+        ReactiveStore(useCase: GetExperienceSummaryUseCase())
+            .getEntities()
+            .map { $0.first?.availableApps }
+            .map { availableApps in
+                guard let availableApps = availableApps else { return false }
```
rh12:
```suggestion
                guard let availableApps else { return false }
```

Core/Core/Features/Profile/Settings/ProfileSettingsViewController.swift
```diff
@@ -326,6 +383,37 @@ public class ProfileSettingsViewController: ScreenViewTrackableViewController {
         return [row]
     }
 
+    private func addCareerExperienceSwitch() async -> Row? {
+        guard
+            AppEnvironment.shared.app == .student,
+            let appExperienceInteractor,
+            await appExperienceInteractor.isExperienceSwitchAvailableAsync() == true else {
```
rh12:
```suggestion
            await appExperienceInteractor.isExperienceSwitchAvailableAsync() else {
```

Core/CoreTests/Features/Experience/APIExperienceSummaryTests.swift
rh12, line 1: Could you please use the same folderstructure here as we have for the production code?
Same for all files here

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
szabinst: I don't recall such thing.

Core/Core/Common/Extensions/CoreData/StringSetTransformer.swift
szabinst: I didn't see it risky but changed to a guard statement just to be extra safe.

Core/Core/Features/Experience/Data/SwitchExperienceUseCase.swift
szabinst: It's not possible because the login process will fetch the summary at least once. Even if - for some reason- there's no entity, the `update` method would simply return early.

ndrsszsz (approved):
QA +1

Nice job! Everything seems to be working fine for me. Tested the switch, login, logout, change user, appearance modes.

Core/Core/Features/Experience/Domain/ExperienceSummaryInteractor.swift
vargaat: Valid concern. Also ReactiveStores in the file don't use the stored environment object.

PR #3558, opened 2025-08-07
[Horizon] ScoresViewModel Rebuilding Fix

reabbotted:
This is a fix to make it so the ScoresViewModel doesn't get rebuilt over and over.

The issue came from the CourseDetailsViewModel. Whenever the HCourse would get updated (which could get notified multiple times in a row), it would cause the ScoresView to rebuild. This ScoresView was calling the ScoresAssembly, which it probably shouldn't anyway.

The fix is to put the ScoresViewModel in the CourseDetailsViewModel. This can get rebuilt if the HCourse ID changes. Because the ScoresViewModel depends on building the ScoresInteractor, I've added a builder parameter to the CourseDetailsViewModel so it can request to have the ScoresViewModel rebuilt as needed.

Additionally, the ScoresInteractor must continually listen for changes to the Course so it can rebuild the scores string if it changes in the repository.

affects: Horizon

### DIFF
--- a/Horizon/Horizon/Sources/Features/Learn/Scores/View/ScoresViewModel.swift
+++ b/Horizon/Horizon/Sources/Features/Learn/Scores/View/ScoresViewModel.swift
@@ -72,7 +72,6 @@ final class ScoresViewModel {
         unowned let unownedSelf = self
 
         selectedFilterOptionRelay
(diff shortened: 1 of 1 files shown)


inst-danger:

</details>

PR #3567, opened 2025-08-11
Followup: Fix (NeedsGrading & Un-submitted) Counts for Submissions Breakdown for Custom Graded ones

suhaibabsi-inst:
refs: MBL-18954
affects: Teacher
release note: none.

This is to defend against the possible case where a submission with needsGrading or un-submitted status is not being counted as "graded" of submission summary API response. And it is not counted as "unsubmitted" or "ungraded" as well. 

## Test Plan

A found use case,
Grade a a student with no submission with "Missing" label, submit that on SpeedGrader, then go back to the same student and assign a custom label to it, then submit again. 
As consequence, Graded count on submissions summary won't include that, it will be also excluded from Ungraded & Un-submitted count. 

Now count for needsGrading and un-submitted upon showing submissions list shouldn't show minus value (-1).

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Assignments/Assignment.swift
+++ b/Core/Core/Features/Assignments/Assignment.swift
@@ -223,7 +223,7 @@ extension Assignment {
                 atState: .submitted, .pending_review, .graded
             )
 
-        needsGradingCount = (item.needs_grading_count ?? 0) - customGradeStatedSubmittedCount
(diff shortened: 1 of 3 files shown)


ndrsszsz (dismissed):
QA +1

The above mentioned issue cannot be fixed without fetching all submissions.

PR #3572, opened 2025-08-12
Hide Individual Recipient Checkbox for Teacher and  Force Bulk Messaging

GaneshNavale:
refs: PFS-25718
    affects: Teacher
    release note: When the restrict_student_access feature flag is enabled in the Teacher app,
        the "Individual Recipient" messaging option is hidden. Messages are always sent as bulk
        messages so recipients cannot see each other
    
test plan:
  - Enable the restrict_student_access feature flag
  - Launch the Canvas Teacher app
  - Navigate to Inbox and start composing a new message
  - Verify that the "Individual Recipient" checkbox/toggle is not visible
  - Send the message and confirm recipients cannot see each other
  - Disable the flag and verify that the checkbox/toggle reappears

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


inst-danger:

</details>

Core/Core/Features/Inbox/ComposeMessage/ViewModel/ComposeMessageViewModel.swift
```diff
@@ -542,6 +548,15 @@ final class ComposeMessageViewModel: ObservableObject {
             .store(in: &subscriptions)
     }
 
+    private func bindStudentAccessRestriction() {
+        studentAccessInteractor
+            .isRestricted()
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] isRestricted in
+                self?.isStudentAccessRestricted = isRestricted && self?.env.app == .teacher
```
rh12: Do we need this check for Teacher app?
Users from Student and Parent apps are allowed to send individual messages for multiple recipients?
Plus including this app-check will mean that `isStudentAccessRestricted` will be always false for other apps, which is a lie and may cause issues if we want to restrict different things later based on it.

GaneshNavale: Good point — this check was added because we currently want to enforce this behavior only for Teacher app users.

When the Student Access feature flag is enabled, the goal is to enforce bulk messaging (BCC) for all recipients, including students, so they cannot see the full recipient list.

In other apps, this feature flag already blocks student-to-student conversations, so there’s no scenario where bulk messaging enforcement would be needed, and this change is not required there.

If needed, I can rename `isStudentAccessRestricted` to something more descriptive, such as `isBulkMessagingEnforced`.

rh12: Thanks for the explanation!
I would still prefer to not include this app-check. Omitting it would break nothing, if I understand you correctly.

Without it we remove some minor complexity (no need to think about or test that case) and we gain some robustness (this class in itself will work properly, regardless of restrictions enforced elsewhere).
The `isStudentAccessRestricted` is fine, it says what it does (if it won't have the app-check)

rh12: As discussed, this restriction is not only not required for Student/Parent apps, but those apps should not be restricted this way.
In that case please rename the property to:
`isBulkMessagingEnforcedBasedOnStudentAccessRestriction`
Please also add the following comment above the line where you set it's value:
```
// Enforcing bulk messaging only for Teacher app, because we want to NOT enforce it for Student/Parent apps at the moment.
// We assume restricted Students won't even be able to message other Students.
```

Related Slack conversation: https://instructure.slack.com/archives/C08TPDPP4D6/p1755083611245239

GaneshNavale: I’ve made the requested changes — please review when you have a moment, Thanks!

rh12:
Thanks for the changes!
One note though, please prefer not amending/rebasing/squashing on an already open PR. Just add a new commit with a meaningful message. This allows for easier reviews. In the end all will be squashed anyway before merging, but the PR itself will be more clear.

PR #3576, opened 2025-08-14
Disable File Uploads for Students

GaneshNavale:
refs: PFS-25737
affects: Student
release note: When the restrict_student_access feature flag is enabled in the Student app,
    then user will not be able to upload files

test plan:
    - Enable the restrict_student_access feature flag
    - Launch the Canvas Student app
    - Navigate to Files and click on the + icon
    - Upload file button should be hidden
    - Disable the flag and verify that the Upload file button reappears

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Files/View/FileList/FileListViewController.swift
+++ b/Core/Core/Features/Files/View/FileList/FileListViewController.swift
@@ -18,6 +18,7 @@
 
 import UIKit
 import Combine
+import CombineSchedulers
 
 public class FileListViewController: ScreenViewTrackableViewController, ColoredNavViewProtocol {
     @IBOutlet weak var emptyImageView: UIImageView!
@@ -79,13 +80,15 @@ public class FileListViewController: ScreenViewTrackableViewController, ColoredN
     private var studentAccessInteractor: StudentAccessInteractor?
     private var subscriptions = Set<AnyCancellable>()
     private var isRestricted = false
+    private var scheduler: AnySchedulerOf<DispatchQueue>!
 
     public static func create(
         env: AppEnvironment,
         context: Context,
         path: String? = nil,
         offlineFileInteractor: OfflineFileInteractor = OfflineFileInteractorLive(),
-        studentAccessInteractor: StudentAccessInteractor? = nil
+        studentAccessInteractor: StudentAccessInteractor? = nil,
+        scheduler: AnySchedulerOf<DispatchQueue> = .main
     ) -> FileListViewController {
         let controller = loadFromStoryboard()
         controller.context = context
@@ -93,6 +96,7 @@ public class FileListViewController: ScreenViewTrackableViewController, ColoredN
         controller.path = path ?? ""
         controller.offlineFileInteractor = offlineFileInteractor
         controller.studentAccessInteractor = studentAccessInteractor
+        controller.scheduler = scheduler
         return controller
     }
 
@@ -103,7 +107,6 @@ public class FileListViewController: ScreenViewTrackableViewController, ColoredN
 
         addButton.accessibilityIdentifier = "FileList.addButton"
         addButton.accessibilityLabel = String(localized: "Add Item", bundle: .core)
-
         editButton.accessibilityIdentifier = "FileList.editButton"
 
         emptyImageView.image = UIImage(named: Panda.FilePicker.name, in: .core, compatibleWith: nil)
@@ -129,10 +132,8 @@ public class FileListViewController: ScreenViewTrackableViewController, ColoredN
 
         studentAccessInteractor?
             .isRestricted()
-            .receive(on: DispatchQueue.main)
-            .sink { [weak self] isRestricted in
-                self?.isRestricted = isRestricted
-            }
+            .receive(on: scheduler)
+            .assign(to: \.isRestricted, on: self, ownership: .weak)
             .store(in: &subscriptions)
 
         colors.refresh()
--- a/Core/Core/Features/Files/View/FileList/FileListViewController.swift
+++ b/Core/Core/Features/Files/View/FileList/FileListViewController.swift
@@ -17,6 +17,7 @@
 //
 
 import UIKit
+import Combine
 
 public class FileListViewController: ScreenViewTrackableViewController, ColoredNavViewProtocol {
     @IBOutlet weak var emptyImageView: UIImageView!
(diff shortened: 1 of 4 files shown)


ChanakyaINST:
QA +1

[
Simulator Screenshot - iPhone 16 Pro - 2025-08-22 at 14 43 50
Simulator Screenshot - iPhone 16 Pro - 2025-08-22 at 14 43 57
](url)

Core/CoreTests/Features/Files/View/FileList/FileListViewControllerTests.swift
```diff
@@ -271,6 +275,54 @@ class FileListViewControllerTests: CoreTestCase {
...
+        controller.viewWillAppear(false)
+
+        let expectation = expectation(description: "Restriction applied")
+
+        // Give time for Combine publisher to deliver on main queue
+        DispatchQueue.main.async {
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 1.0)
```
vargaat: A less flaky approach is to not use the main queue in tests but rather than a synchronous scheduler. Search for `AnySchedulerOf<DispatchQueue>` and `DispatchQueue.test` on how to use it.

Core/Core/Features/Files/View/FileList/FileListViewController.swift
```diff
@@ -116,6 +127,14 @@ public class FileListViewController: ScreenViewTrackableViewController, ColoredN
             self.tableView.contentOffset.y = self.searchBar.frame.height
         }
 
+        studentAccessInteractor?
+            .isRestricted()
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] isRestricted in
+                self?.isRestricted = isRestricted
+            }
+            .store(in: &subscriptions)
```
vargaat: In rare cases it could happen that the cache invalidates before moving to this page so the `isRestricted()` method makes an API call that takes time to complete. The screen can load during this time for the user with `isRestricted` being `false`, allowing the user to upload a file.
A better approach would be to keep the loading view on screen until we receive a response from the `isRestricted()` method call. Another solution would be to make the `isRestricted()` call to be synchronous by using a cached value but that would require to load the state of this flag at app start.

Core/CoreTests/Features/Files/View/FileList/FileListViewControllerTests.swift
```diff
@@ -271,6 +275,54 @@ class FileListViewControllerTests: CoreTestCase {
         XCTAssertEqual(controller.tableView.numberOfRows(inSection: 0), 0)
     }
 
+    func test_addFileButtonVisible_whenNotRestricted() {
+        controller = FileListViewController.create(
```
vargaat, line 284: I think in such cases it's better to explicitly declare the initial state of the test.

```suggestion
        studentAccessInteractor.setRestricted(false)

        controller = FileListViewController.create(
```

Core/Core/Features/Files/View/FileList/FileListViewController.swift
```diff
@@ -116,6 +127,14 @@ public class FileListViewController: ScreenViewTrackableViewController, ColoredN
             self.tableView.contentOffset.y = self.searchBar.frame.height
         }
 
+        studentAccessInteractor?
+            .isRestricted()
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] isRestricted in
+                self?.isRestricted = isRestricted
+            }
```
vargaat: nit: Instead of sink and weak self:
```suggestion
            .assign(to: \.isRestricted, on: self, ownership: .weak)
```

vargaat (changes requested):

Core/Core/Features/Files/View/FileList/FileListViewController.swift
GaneshNavale: Thanks for the suggestion! The backend already blocks file uploads for restricted users, so even if the button appears, the user can’t actually upload. I also checked `loadingView.isHidden` during the isRestricted assignment, and it’s already true, so the loading state behaves as expected. I can add an extra check if you think it’s needed, but the current approach should be sufficient.

Core/Core/Features/Files/View/FileList/FileListViewController.swift
GaneshNavale: Done

Core/CoreTests/Features/Files/View/FileList/FileListViewControllerTests.swift
GaneshNavale: Done

Core/CoreTests/Features/Files/View/FileList/FileListViewControllerTests.swift
GaneshNavale: Done

Core/CoreTests/Features/Files/View/FileList/FileListViewControllerTests.swift
```diff
@@ -271,6 +278,51 @@ class FileListViewControllerTests: CoreTestCase {
...
+            studentAccessInteractor: studentAccessInteractor,
+            scheduler: testScheduler.eraseToAnyScheduler()
+        )
+
+        controller.view.layoutIfNeeded()
+        controller.viewWillAppear(false)
+
+        // Move virtual time forward by 100ms
+        testScheduler.advance(by: .seconds(0.1))
+
```
vargaat, line 318: Thanks, this way the test is more deterministic. Since we are not really interested in how much time have elapsed, we can use az immediate scheduler to further simplify the test and remove this advance call.

```
    private var testScheduler: AnySchedulerOf<DispatchQueue>!

    override func setUp() {
        super.setUp()
        testScheduler = DispatchQueue.immediate.eraseToAnyScheduler()
```
```suggestion
```

vargaat (approved):
Looks good! If you want updated builds please merge master and add the `builds:` line to the PR's description. See details [here](https://github.com/instructure/canvas-ios/pull/3578).

Core/Core/Features/Files/View/FileList/FileListViewController.swift
vargaat: If the API already blocks file uploads for these users then we can ignore this edge case.

Core/CoreTests/Features/Files/View/FileList/FileListViewControllerTests.swift
GaneshNavale: Thanks, I will commit these suggestions in the next pull request.

PR #3581, opened 2025-08-18
Fix Parent E2E test crash

vargaat:
There was a force unwrap that crashed if the view could not be found on screen on the first try.

refs: MBL-19097
affects: none
release note: none

test plan:
- Run test AddStudentTests and ManageStudentsTests locally as on CI the test suite never finishes due to the beta environment being unstable.


rh12 (approved):
QA + 1, both tests succeeded locally

PR #3583, opened 2025-08-20
Resolve Consortia Root User issue of Inaccessible Classic Quizzes in Certain Menus

suhaibabsi-inst:
refs: MBL-19108
affects: Student, Teacher
builds: Student, Teacher
release note: Fixed issue of shard account user not being able to access classic quizzes from certain menus

## What changed

The solution was about inventing a mechanism to distinguish between account shardID of the user logged in, from that attached to courses owned by other account. For assignment/quiz & submission details, the global ID of course must be constructed of that `shardID` of the owning account, in order for APIs of course, assignment/quiz and submission fetching to work properly.

Property `asRoot` was tweaked to reflect this new understanding, thus providing the option to pass the exact `shardID` to use to construct the full ID of a course, assignment/quiz or user.

Another important change was about passing the information of `shardID` for a course, to the `router` so to customize the creation of environment (override) passed to route handlers. 

A new method was introduced to `CourseTabUrlInteractor` to figure out course `shardID` based on baseURL overrides map if necessary. Based on the assumption that accessing a URL of different baseURL than the one user logged-in would only occurs, in mobile context, if he's about to navigate through inner pages a course owned by other account. 

The logic was at first trying to figure out course `shardID` from URL itself, if not available, we attempt to fetch a complete courseID from baseURLOverrides map of the interactor. See `CourseTabUrlInteractor.courseShardID(for:)` for more details.

## Test Plan

- See [ticket's](https://instructure.atlassian.net/browse/MBL-19108) description.
- Also, I did a regression testing on previous Consortia root-vs-child account issues so to make sure nothing is broken along the way. I'd recommend doing that as well. 
- Finally a smoke test on the changed screens (course details, assignment/quiz details, pages, modules) would be helpful as well. 

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [x] Tested on tablet
- [ ] Tested in dark mode
- [x] Tested in light mode
- [ ] Approve from product


inst-danger:
<h3>Builds</h3><p><strong>Commit:</strong> Merge branch &#39;master&#39; into bugfix/MBL-19108-Cross-Shard-Classic-Quizzes-In-Certain-Menus (<a href="https://github.com/instructure/canvas-ios/commit/403da47">403da47</a>)<br/><strong>Build Number:</strong> 141<br/><strong>Built At:</strong> Aug 27 13:07 CEST (08/27 05:07 AM MDT)</p><table><tr><td valign="top" width="250" align="center"><details open><summary> Student</summary><a href="https://app.bitrise.io/app/5a9c92bb-8a36-489b-85d9-8665096f5127/installable-artifacts/e125cb62a8d9c973/public-install-page/0ab2850881af98edfb6b214df6189784"></a></details></td><td valign="top" width="250" align="center"><details open><summary> Teacher</summary><a href="https://app.bitrise.io/app/5a9c92bb-8a36-489b-85d9-8665096f5127/installable-artifacts/b5d11cf44176fc03/public-install-page/b2608b6da273c3adbcded115e0729476"></a></details></td></tr></table>

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
```diff
@@ -71,6 +71,9 @@ open class AppEnvironment {
     open var root: AppEnvironment { self }
     public var isRoot: Bool { root === self }
 
+    public var sessionShardID: String? { currentSession?.accessToken?.shardID }
+    public var courseShardID: String? { sessionShardID }
```
rh12: I may miss something, but it seems it's the exact same value used in both computed properties.
Or is this groundwork for some future change?

Core/Core/Common/CommonModels/Router/CourseTabUrlInteractor.swift
```diff
@@ -239,6 +239,36 @@ public final class CourseTabUrlInteractor {
...
+            return shardID
+        }
+
+        if let urlHost = url.host,
+           let shardID = baseURLHostOverridesPerCourse
+            .first(where: { $0.key.id.hasShardID && $0.value == urlHost })?
+            .key
+            .id
+            .shardID {
+            return shardID
```
rh12:
```suggestion
        if let urlHost = url.host,
           let overrideContext = baseURLHostOverridesPerCourse.first(where: { $0.value == urlHost })?.key,
           let shardID = overrideContext.id.shardID {
            return shardID
        }
```

I think this is a bit more clear, and it allows to check the value first, and if it's the actual override we are looking for, only then check the shardID (`hasShardID` seemed to be redundant plus it would recompute `shardID` needlessly again)

Same for the other `if` below

Core/Core/Common/CommonModels/Router/CourseTabUrlInteractor.swift
```diff
@@ -239,6 +239,36 @@ public final class CourseTabUrlInteractor {
             return $0.key.isEquivalent(to: context)
         })?.value
     }
+
+    public func courseShardID(for url: URLComponents) -> String? {
+        let context = Context(path: url.path)
```
rh12:
```suggestion
        let pathContext = Context(path: url.path)
```
For clarity, along with my other comment.

Core/Core/Features/Courses/GetCourses.swift
```diff
@@ -35,9 +35,9 @@ public class GetCourse: APIUseCase {
         let modifiedCase = self
 
         if env.isRoot == false {
-            // Should always be used in root-form ID when this
+            // Should always be used in global-form ID when this
             // use case is used in inner course details pages
```
rh12, line 39: I think I don't really understand this comment.
Do you mean that this use case should always use the global ID format?
Or only in some cases ("inner course details pages")?
If the later, shouldn't we have a parameter for this?

Core/Core/Features/Modules/ModuleItems/ModuleItemSequenceViewController.swift
```diff
@@ -111,7 +111,9 @@ public final class ModuleItemSequenceViewController: UIViewController {
         guard let url = url.url else { return nil }
         if let current = sequence?.current {
             return ModuleItemDetailsViewController.create(env: env, courseID: courseID, moduleID: current.moduleID, itemID: current.id)
-        } else if assetType != .moduleItem, let match = env.router.match(url.appendingOrigin("module_item_details")) {
+        } else if
+            assetType != .moduleItem,
+            let match = env.router.match(url.appendingOrigin("module_item_details")) {
```
rh12: If you want to break it:
```suggestion
        } else if assetType != .moduleItem,
                  let match = env.router.match(url.appendingOrigin("module_item_details")) {
```

Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
```diff
@@ -172,7 +173,8 @@ class SubmissionButtonPresenter: NSObject {
     }
 
     private func presentStudentAnnotation(assignment: Assignment, view: UIViewController) {
-        let courseScope = Scope(predicate: NSPredicate(format: "%K == %@", #keyPath(Course.id), assignment.courseID), order: [])
+        let courseID = assignment.courseID.asNonRootPrefixedCourseID(in: env)
```
rh12: I'm not sure how to handle this, but in general my problem is that these kind of id transformations seem kind of random. Why is this needed for StudentAnnotations but not for else?
Maybe a comment explaining this would help.
Or rather if this is something that is relevant for each student annotation opened from anywhere, maybe the `Assignment` itself could have a `courseIdForSubmissionType()` or similar and that could decide based on the type of submission. Of course it's still something we need to remember to use elsewhere, but at least the logic would be closer to the issue.
Not sure if this makes sense, I'm not that well versed in this area, just aiming to somehow centralize all these (seemingly) random ID transformations and variants a bit, if possible.

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
suhaibabsi-inst: `courseShardID` is meant to be overridden in `AppEnvironmentOverride`

Core/Core/Common/CommonModels/Router/CourseTabUrlInteractor.swift
suhaibabsi-inst: Strange enough, I have tested a case where some of the keys which are representative of Course Context, might have a duplicate value for the same course, one with sharedID and another without. 🙂

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
rh12: Ah, got it, thanks! Could you please add a short code comment for this?

Core/Core/Features/Courses/GetCourses.swift
suhaibabsi-inst: By checking `isRoot`, you test whether this environment has been passed as the root one or not.

Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
suhaibabsi-inst: This is primarily to serve the purpose of fetching course using the correct ID format. 
Fetching courses use cases like `GetCourse` & `GetCourses` actually receive courses with IDs of such format only when called in a cross-shard setup. For regular case, it is received in the local format. That's why this logic is necessary here to match IDs format across all cases. 
I will add a comment to explain that.

rh12: I see, so this is not dependent on the submission type, but it is something that all such usecases should do.
So basically this presenter should use a usecase, and the logic should be in usecase's `modified()` method?
(not saying to refactor it now, just want to understand)

What I don't get now is why is the `if env.isRoot { return localID }` in the property used here, and not in `GetCourse.modified()`? I've tried to see whether we handle this elsewhere, but found nothing like it.

suhaibabsi-inst: The modified method of use case, is meant to prepare the use case for both API request, and scoping of local objects as well. And it can be used as necessary. And the thinking behind it is to provide a place for local reasoning about how use case properties like IDs should be treated for both request stages of the use case (from API, or from DB). 

I know this would seem to complicate things for a bit, but I would expect it to evolve to be like a central source of truth about such properties. And we can definitely think about a refactor for it when we reach to a stable rationale about it later on.

For the other method, this meant primarily to construct the course ID proper for the current point of fetching the course from database, which occurs in Line 186 after it. And the logic assumes that if we reached to this inner page with a root environment, that means we are no longer in the cross-shard reality. Thus the expectation is to have it in the local format. It is like we are setting our expectation for it explicitly here.

suhaibabsi-inst: Ahh, I got your comment now.
We can get rid of Line 186, and replace it with `GetCourse` use case, and the `modified` method would do the necessary work.
This way, we can get rid of this method 🙂

Core/Core/Common/CommonModels/Router/CourseTabUrlInteractor.swift
rh12: That's weird. Could you please then update the logic accordingly and make this hidden knowledge explicit?
Something like this (didn't test it)
```
        if let urlHost = url.host {
            let overrideContexts = baseURLHostOverridesPerCourse
                .filter { $0.value == urlHost } // It's possible to have multiple contexts with the same `value`
                .map { $0.key }
            if let shardID = overrideContexts.compactMap(\.id.shardID).first {
                return shardID
            }
        }
```
And also for the other `if` below.

Please also cover these with unit tests to make sure we won't refactor these edge cases away.

Student/Student/Submissions/SubmissionButton/SubmissionButtonPresenter.swift
rh12: Sound great, less similar stuff for almost-the-same-stuff-but-slightly-different-stuff is the better :)
Thank you!

Core/Core/Common/CommonModels/Router/CourseTabUrlInteractor.swift
rh12: Could you please still break these two `if`s up for better readability?

Core/Core/Common/CommonModels/Store/UseCase.swift
```diff
@@ -119,6 +119,12 @@ public extension UseCase {
         }
     }
 
+    /// Execute fetch request synchronously from local database
+    func fetchFromDatabase(environment: AppEnvironment = .shared) -> [Model] {
+        let modifiedUseCase = modified(for: environment)
```
vargaat, line 126: Let's add an assertion here to check if this method was called on the main thread. The view context is associated with the main thread so if we call this from a background thread the app will randomly crash.

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
```diff
@@ -71,6 +71,12 @@ open class AppEnvironment {
     open var root: AppEnvironment { self }
     public var isRoot: Bool { root === self }
 
+    public var sessionShardID: String? { currentSession?.accessToken?.shardID }
+
+    // This can be different from `sessionShardID` for cross-shard setup.
+    // See `AppEnvironmentOverride` impl. for it.
+    public var courseShardID: String? { sessionShardID }
```
vargaat, line 78: Can we have groups here? If the user is part of a course group that is from a child instance, then entering the group from the dashboard should create an environment override with the group's shard id here. If this is possible, then the name selection with the course word in it might not be the best. `contextShardID` maybe?

suhaibabsi-inst: I think we better introduce that in a separate work item, as it requires we have some major changes on code of Groups UI, mainly in `GroupNavigationViewController`, in order to utilize this. Which could broaden the scope of this bug fix by far. Here is the ticket created to update that. And I will be updating the ticket to have some test accounts for this specific use case:
https://instructure.atlassian.net/browse/MBL-19244

vargaat: Agree, we can change the name later.

vargaat (approved):

PR #3594, opened 2025-08-25
Fixed issue of Un-viewable Studio Embeds in cross-shard instances

suhaibabsi-inst:
refs: MBL-19100
affects: Student
builds: Student
release note: Fixed issue of Unviewable Studio Embeds in corss-shard instances

Basically the fix was to reset cookies on the web view used in assignment details screen, for the child domain with same values used for the root one. The core of this fix can be found in the private method `CoreWebView. copyRootDomainCookies(_:)`

## Test Plans

See [ticket's](https://instructure.atlassian.net/browse/MBL-19100) description.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


rh12 (commented):
Code + 1

suhaibabsi-inst:
> The solution works for the issue but it feels weird. Webviews use shared cookie stores and the `cookieKeepAliveTimer` ensures that session cookies in this store are up-to-date for the logged-in domain so any webview opened in the app will have a valid session. Your solution moves all such cookies to another domain and I have a feeling that other screens that are using the logged-in domain won't work after this, until the keep alive fires and restores them. My idea would be to seed the webview with the session cookies for the child domain, before we load the actual html snippet. Check the code in the cookie keep alive logic. Another idea would be not to modify the original cookies but to duplicate them with the new domain.

The `setCookie(_:)` method according to [this original documentation](https://developer.apple.com/documentation/webkit/wkhttpcookiestore/setcookie(_:completionhandler:)), do replacement only when name, domain and path of the passed one matches with one of those saved in the storage. Otherwise, it will add that to the storage. 

So, for the current implementation we actually **_duplicate_** cookies with `domain` property value of the root one, altering this property to match the domain of child one. And I did try that out to make sure nothing is broken with the original cookies of the root one.

Let me know your thoughts given this new understanding.

Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift
```diff
@@ -400,6 +403,46 @@ open class CoreWebView: WKWebView {
         guard contentHeight.isFinite else { return }
         sizeDelegate?.coreWebView(self, didChangeContentHeight: contentHeight)
     }
+
+    public func resetEnvironment(_ env: AppEnvironment) {
+        self.env = env
+        self.resetCookiesForEnvironmentOverride(env)
+    }
+
+    private func resetCookiesForEnvironmentOverride(_ env: AppEnvironment) {
```
vargaat: I just realized that this is an async process used in a synchronous way and it could happen that setting cookies do not finish before we try to load the webview content. Please try to guarantee that this method finishes before the webview starts to load content.

Core/Core/Common/CommonUI/CoreWebView/View/CoreWebView.swift
```diff
@@ -400,6 +403,57 @@ open class CoreWebView: WKWebView {
...
+        else {
+            completion?()
+            return
+        }
+
+        configuration
+            .websiteDataStore
+            .httpCookieStore
+            .getAllCookies()
+            .receive(on: DispatchQueue.main)
```
vargaat: If we expect the `completion` callback on the main thread then we should add this to after the `flatMap` because that can replace the used thread. I'd guess cookies can be used from background threads so I'm not sure if we need to process them on the main queue. Since this is a quite complex method affecting a core component, let's add some unit tests to it.

vargaat (changes requested):

ndrsszsz (approved):
QA +1

vargaat (approved):
Another shard related PR was merged into this.

PR #3607, opened 2025-08-28
[Student] DCP - Update todo list and todo widget

ndrsszsz:
## DCP - Update todo list and todo widget

refs: MBL-19134
builds: Student, Teacher
affects: Student
release note: Upgraded todo screen and widget to support Discussion Checkpoints.
test plan: See ticket.

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -0,0 +1,62 @@
...
+    var todosPublisher: AnyPublisher<[TodoItem], Error> { get }
+}
+
+public final class TodoInteractorLive: TodoInteractor {
+    public let todosPublisher: AnyPublisher<[TodoItem], Error>
+
+    init(env: AppEnvironment, startDate: Date, endDate: Date) {
+        self.todosPublisher = ReactiveStore(useCase: GetAllCourses())
+            .getEntities()
+            .compactMap {
```
vargaat: Since we don't return nil values from this block a simple `map` will do the job.

Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Foundation
+import Combine
+
+public protocol TodoInteractor {
+    var todosPublisher: AnyPublisher<[TodoItem], Error> { get }
```
vargaat: I'd suggest to create a `CurrentValueSubject` that stores cached todo items here and a separate `refresh(ignoreCache: Bool) -> AnyPublisher<Void, Error>` method that does the download. This way you can support pull to refresh with the cache ignored and the regular download using cache. Also testing will be easier because the interactor will do nothing until you call the refresh method.

Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    init(env: AppEnvironment, startDate: Date, endDate: Date) {
+        self.todosPublisher = ReactiveStore(useCase: GetAllCourses())
+            .getEntities()
+            .compactMap {
+                var contextCodes: [String] = $0.map { $0.canvasContextID }
+                if let userContextCode = Context(.user, id: env.currentSession?.userID)?.canvasContextID {
+                    contextCodes.append(userContextCode)
+                }
+                return contextCodes
```
vargaat, line 54: On my account I receive a 403 for these context codes. I think one of my courses is unpublished maybe and that causes the issue but I remember facing this previously, adding a simple filter here will solve it.

Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -0,0 +1,62 @@
...
+            }
+            .flatMap { codes in
+                return ReactiveStore(useCase: GetPlannables(startDate: startDate, endDate: endDate, contextCodes: codes))
+                    .getEntities(ignoreCache: true)
```
vargaat: The previous implementation exhausted, we should add `loadAllPages: true` if we don't have paging implemented on the UI.

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -0,0 +1,61 @@
...
+                    }
+
+                    Text(item.date.dateTimeStringShort)
+                        .font(.regular14)
+                        .foregroundStyle(.textDark)
+                }
+
+                Spacer()
+
+                Image(systemName: "chevron.right")
```
vargaat: Can we use `InstUI.DisclosureIndicator`?

Core/Core/Features/Todos/View/TodoListScreen.swift
```diff
@@ -0,0 +1,96 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import SwiftUI
+
+public struct TodoListScreen: View {
+    @Environment(\.viewController) private var viewController
+    @ObservedObject var viewModel: TodoListViewModel
```
vargaat, line 24: We should add `@Environment(\.dynamicTypeSize) private var dynamicTypeSize` to make the screen redraw when font size is increased. Probably we need to add this to the cell as well.

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+    @Published var items: [TodoItem] = []
+    @Published var hasError: Bool = false
+
+    private let env: AppEnvironment
+    private let interactor: TodoInteractor
+    private var subscriptions = Set<AnyCancellable>()
+
+    init(env: AppEnvironment = .shared, interactor: TodoInteractor? = nil) {
+        self.env = env
+        self.interactor = interactor ?? TodoInteractorLive(env: env, startDate: .now.startOfDay(), endDate: .now.addDays(28))
```
rh12: 1. Could we move this magic number into a constant and explain it a bit?
2. Shouldn't it be the endOfDay() there?
3. IMHO this should be defined in the interactor to ensure all future entry points show the same interval. (or update it later if we explicitly want to use different intervals)
4. Please inject the interactor from the assembly, and remove the optionality here

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+public class TodoListViewModel: ObservableObject {
+
+    @Published var items: [TodoItem] = []
+    @Published var hasError: Bool = false
+
+    private let env: AppEnvironment
+    private let interactor: TodoInteractor
+    private var subscriptions = Set<AnyCancellable>()
+
+    init(env: AppEnvironment = .shared, interactor: TodoInteractor? = nil) {
```
rh12: 1. Please swap `env` parameter with `interactor` (also the properties)
2. env should also not have a default value, we want to enforce env propagation.

Core/Core/Features/Todos/TodoAssembly.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+
+import SwiftUI
+
+public struct TodoAssembly {
+    public static func makeTodoListViewController(env: AppEnvironment = .shared) -> UIViewController {
+        let model = TodoListViewModel(env: env)
+        let todoVC = CoreHostingController(TodoListScreen(viewModel: model))
+        todoVC.navigationBarStyle = .global
+        todoVC.navigationItem.titleView = Brand.shared.headerImageView()
```
rh12, line 27: IMHO these belong to the TabBarController. If we have other entry points later probably don't want this to look global.

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -0,0 +1,61 @@
...
+            HStack(alignment: .center) {
+                VStack(alignment: .leading, spacing: 4) {
+                    Text(item.contextName)
+                        .font(.regular14)
+                        .foregroundStyle(item.color)
+
+                    Text(item.title)
+                        .font(.regular16)
+                        .foregroundStyle(.textDarkest)
+                        .lineLimit(2)
```
rh12: I don't think we should have a line limit here, please align this with Marci

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -0,0 +1,61 @@
...
+                    Text(item.date.dateTimeStringShort)
+                        .font(.regular14)
+                        .foregroundStyle(.textDark)
+                }
+
+                Spacer()
+
+                Image(systemName: "chevron.right")
+                    .font(.regular14)
+                    .foregroundStyle(.textDarkest)
```
rh12:
```suggestion
                            InstUI.DisclosureIndicator()
                                .paddingStyle(.leading, .cellAccessoryPadding)
```

Core/Core/Features/Todos/View/TodoListScreen.swift
```diff
@@ -0,0 +1,96 @@
...
+                        TodoListItemView(item: item)
+                            .paddingStyle(set: .iconCell)
+                    }
+                    InstUI.Divider(item != viewModel.items.last ? .padded : .full)
+                }
+            }
+        } refreshAction: { completion in
+            viewModel.refresh(completion: completion)
+        }
+        .padding(.vertical, 8)
```
rh12: I don't think we should have vertical paddings here. Maybe at the bottom, but probably not even that

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -0,0 +1,61 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import SwiftUI
+
+struct TodoListItemView: View {
```
rh12: Could we please build this like the cells in InstUI?
Include Button, paddings, a11y, etc. here.
And rename it to `TodoListItemCell`

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -0,0 +1,61 @@
...
+    let item: TodoItem
+
+    var body: some View {
+        HStack(alignment: .top, spacing: 16) {
+            item.icon
+                .font(.regular14)
+                .foregroundStyle(item.color)
+                .frame(width: 25)
```
rh12: This should use `.scaledIcon` instead of setting frame

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -0,0 +1,61 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import SwiftUI
+
+struct TodoListItemView: View {
+    let item: TodoItem
+
+    var body: some View {
+        HStack(alignment: .top, spacing: 16) {
```
rh12: All this should be a11y combined probably, with the icons hidden

Core/Core/Features/Todos/View/TodoListScreen.swift
```diff
@@ -0,0 +1,96 @@
...
+    @ViewBuilder
+    private var contentView: some View {
+        if viewModel.hasError {
+            errorView
+        } else if viewModel.items.isEmpty {
+            emptyView
+        } else {
+            dataView
+        }
+    }
```
rh12: Couldn't we replace all this with BaseScreen?
Maybe the sleeping panda will block that, but I would even sacrifice that for now and add support for non-moving pandas in the future

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+        switch item.type {
+        case .planner_note:
+            let vc = PlannerAssembly.makeToDoDetailsViewController(plannableId: item.id)
+            env.router.show(vc, from: viewController, options: .detail)
+        case .calendar_event:
+            let vc = PlannerAssembly.makeEventDetailsViewController(eventId: item.id)
+            env.router.show(vc, from: viewController, options: .detail)
+        default:
+            guard let url = item.htmlURL else { return }
+            let to = url.appendingOrigin("calendar")
```
rh12:
```suggestion
            let to = url.appendingOrigin("todo")
```

Core/Core/Features/Todos/Model/TodoItem.swift
```diff
@@ -17,78 +17,81 @@
...
+        id: String = "1",
         type: PlannableType = .assignment,
         date: Date = Clock.now,
         title: String = "Example Assignment",
+        subtitle: String = "Subtitle",
         contextName: String = "Example Course",
         htmlURL: URL? = nil,
         color: Color = .red,
         icon: Image? = Image.assignmentLine
     ) -> TodoItem {
```
rh12: Could we please don't add random default values here? We can add values in tests if and when they are needed. Just default to `nil`, `""` where possible

rh12 (changes requested):

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: Done.

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: Agreed. Done.

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: Thanks. Unfortunately filtering for isPublished doesn't solve the issue.

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: Thanks. Done.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Sure, thanks. Done.

Core/Core/Features/Todos/View/TodoListScreen.swift
ndrsszsz: Done.

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
ndrsszsz: 1. Actually, that magic number is not even needed anymore.
2. It should be distantFuture (we should get all items from now to infinity).
3. Agreed, done.
4. Agreed, done.

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
ndrsszsz: Agreed, done.

Core/Core/Features/Todos/TodoAssembly.swift
ndrsszsz: Let's discuss this on slack, please.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Thanks. Asked him.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Thanks, Attila already mentioned this. Done.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Done.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Let's discuss this on slack, please.

Core/Core/Features/Todos/View/TodoListScreen.swift
ndrsszsz: Let's discuss this on slack, please.

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
ndrsszsz: Done.

Core/Core/Features/Todos/Model/TodoItem.swift
ndrsszsz: Not agreed but understood. Done.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: He said we can remove it, so I did.

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: But changing the useCase from GetAllCourses to GetCourses did. Pushed the change.

Core/Core/Features/Todos/View/TodoListItemView.swift
```diff
@@ -19,14 +19,15 @@
...
 
 struct TodoListItemView: View {
+    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
+
     let item: TodoItem
 
     var body: some View {
         HStack(alignment: .top, spacing: 16) {
-            item.icon
+            item.icon?.scaledIcon()
```
rh12: This should not be an optional, we always return a value there. Please update the `Plannable.icon` type.
And please break the modifier to it's own line

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
```diff
@@ -24,18 +24,20 @@ public class TodoListViewModel: ObservableObject {
...
-        self.interactor = interactor ?? TodoInteractorLive(env: env, startDate: .now.startOfDay(), endDate: .now.addDays(28))
+
         refresh()
     }
 
-    public func refresh(completion: (() -> Void)? = nil) {
-        interactor.todosPublisher
+    public func refresh(ignoreCache: Bool = false, completion: (() -> Void)? = nil) {
+        interactor.refresh(ignoreCache: ignoreCache)
+        interactor.todos
```
rh12: I think this subscription should be done in init, once.

Core/Core/Features/Todos/TodoAssembly.swift
ndrsszsz: Stays as is, as discussed.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Done.

Core/Core/Features/Todos/View/TodoListScreen.swift
ndrsszsz: Done.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Done.

Core/Core/Features/Todos/View/TodoListItemView.swift
ndrsszsz: Done.

Core/Core/Features/Todos/ViewModel/TodoListViewModel.swift
ndrsszsz: Done.

rh12 (commented):
Profile menu button is missing, otherwise QA + 1

Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -0,0 +1,84 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Foundation
+import Combine
+
+public protocol TodoInteractor {
+    var todos: AnyPublisher<[TodoItem], Never> { get }
+
+    func refresh(ignoreCache: Bool) -> AnyPublisher<Bool, Error>
```
vargaat: We could introduce a `typealias` for `Bool` named `IsEmptyState` so the type `AnyPublisher<IsEmptyState, Error>` would be more descriptive.

Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    init(startDate: Date = .now, endDate: Date = .distantFuture, env: AppEnvironment) {
+        self.startDate = startDate
+        self.endDate = endDate
+        self.env = env
+    }
+
+    public func refresh(ignoreCache: Bool) -> AnyPublisher<Bool, Error> {
+        ReactiveStore(useCase: GetCourses())
+            .getEntities()
```
vargaat: `ignoreCache` should be forwarded here.

Core/CoreTests/Features/Todos/Model/TodoInteractorLiveTests.swift
```diff
@@ -0,0 +1,239 @@
...
+        XCTAssertFirstValueAndCompletion(testee.refresh(ignoreCache: false)) { isEmpty in
+            XCTAssertFalse(isEmpty)
+        }
+        XCTAssertFirstValue(testee.todos) { todos in
+            XCTAssertEqual(todos.count, 1)
+            XCTAssertEqual(todos[0].title, "Assignment 1")
+        }
+    }
+
+    func testRefreshWithIgnoreCache() {
```
vargaat, line 112: A proper test here would be to make a refresh with `ignoreCache` `false` and then with `true` values and check if the second call triggers all API calls or not. We have mock methods that accept a block that gets invoked when the API is invoked, so you know if the API was called or not. If we don't want to spend time on this check, then let's just delete this test, because it doesn't test if the `ignoreCache` parameter is passed to all subsequent refreshes or not.

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: Done.

Core/Core/Features/Todos/Model/TodoInteractor.swift
ndrsszsz: Done.

Core/CoreTests/Features/Todos/Model/TodoInteractorLiveTests.swift
ndrsszsz: Done.

PR #3609, opened 2025-08-29
SpeedGrader - Show comments before first attempt

vargaat:
refs: [MBL-19225](https://instructure.atlassian.net/browse/MBL-19225)
builds: Teacher
affects: Teacher
release note: Updated SpeedGrader to show student comments made before the first submission.

test plan:
- Enable assignment enhancements.
- Create a submission.
- As a student add a comment0 to it.
- Make the 1st submission.
- Add comment1.
- Make a 2nd submission.
- Add comment2.
- Open SpeedGrader.
- 1st attempt should display comment0 and comment1 while 2nd attempt should display only attempt2.

## Checklist

- [x] Tested on phone
- [ ] Tested on tablet

[MBL-19225]: https://instructure.atlassian.net/browse/MBL-19225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Teacher/Teacher/SpeedGrader/Comments/Model/SubmissionCommentFilter.swift
+++ b/Teacher/Teacher/SpeedGrader/Comments/Model/SubmissionCommentFilter.swift
@@ -43,13 +43,9 @@ struct SubmissionCommentFilterLive: SubmissionCommentFilter {
             let commentAttempt = comment.attemptFromAPI?.intValue
             let isCommentBeforeFirstAttempt = commentAttempt == 0
 
-            if commentAttempt == nil
+            return commentAttempt == nil
                 || commentAttempt == attempt
-                || (attempt == 1 && isCommentBeforeFirstAttempt) {
-                return true
-            }
-
-            return false
+                || (attempt == 1 && isCommentBeforeFirstAttempt)
         }
     }
 }
--- a/Teacher/TeacherUnitTests/SpeedGrader/Comments/ViewModel/SubmissionCommentListViewModelTests.swift
+++ b/Teacher/TeacherUnitTests/SpeedGrader/Comments/ViewModel/SubmissionCommentListViewModelTests.swift
@@ -136,7 +136,7 @@ class SubmissionCommentListViewModelTests: TeacherTestCase {
     // MARK: - filtering
 
     func test_commentFilter_callsProtocolMethodWithCorrectParameters() {
-        let mockFilter = MockSubmissionCommentFilter()
+        let mockFilter = SubmissionCommentFilterMock()
         let comment1 = makeComment(id: "1", attempt: nil)
         let comment2 = makeComment(id: "2", attempt: 1)
         interactor.getCommentsResult = Publishers.typedJust([comment1, comment2])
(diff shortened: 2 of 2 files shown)


Teacher/TeacherUnitTests/SpeedGrader/Comments/ViewModel/SubmissionCommentListViewModelTests.swift
```diff
@@ -348,3 +332,17 @@ class SubmissionCommentListViewModelTests: TeacherTestCase {
         )
     }
 }
+
+private class MockSubmissionCommentFilter: SubmissionCommentFilter {
```
rh12: nit, for consistency with Interactor namings
```suggestion
private class SubmissionCommentFilterMock: SubmissionCommentFilter {
```

Teacher/Teacher/SpeedGrader/Comments/ViewModel/SubmissionCommentListViewModel.swift
```diff
@@ -129,21 +136,15 @@ class SubmissionCommentListViewModel: ObservableObject {
         selectedAttemptNumber = attempt
         guard state == .data || state == .empty else { return }
 
-        let comments = filterComments(allComments, for: attempt)
+        let comments = commentFilter.filterComments(
+            allComments,
+            for: attempt,
+            isAssignmentEnhancementsEnabled: isAssignmentEnhancementsEnabled
```
rh12, line 142: IMHO it would be cleaner to pass this in during init, as this is supposed to be basically constant.
It would mean we could create the filter only during load, and maybe make it optional.
If it stays like this, that's also viable, just my 2cents

Teacher/Teacher/SpeedGrader/Comments/Model/SubmissionCommentFilter.swift
```diff
@@ -0,0 +1,55 @@
...
+            let commentAttempt = comment.attemptFromAPI?.intValue
+            let isCommentBeforeFirstAttempt = commentAttempt == 0
+
+            if commentAttempt == nil
+                || commentAttempt == attempt
+                || (attempt == 1 && isCommentBeforeFirstAttempt) {
+                return true
+            }
+
+            return false
```
rh12: We could simply return the expression

Teacher/Teacher/SpeedGrader/Comments/ViewModel/SubmissionCommentListViewModel.swift
vargaat: If the property is optional then it adds complexity. If we remove the flag from the method then we need to add it back to somewhere in the protocol (property, init?) and this makes it harder to understand. In this format it's pretty clear that the function's purpose is to do the filtering using all these inputs.

vargaat:
> Following the test plan mentioned, and turned on assignments enhancements for the account I am using. But I still see all comments in SpeedGrader comments section, regardless of selected attempt. See following:

I updated the test plan. It seems the flag was off on your course's level and you made comments without the flag being enabled so they didn't get assigned to an attempt.

PR #3613, opened 2025-08-29
refs: CLX-2815

reabbotted:
refs: CLX-2815
affects: Student
release note: Fixing a bug
builds: student

The region was not getting set on the session when a user logged in so when DomainService went to construct the URL, it was always using east for the region.

### DIFF
--- a/Horizon/HorizonTests/Common/Data/API/DomainServiceTests.swift
+++ b/Horizon/HorizonTests/Common/Data/API/DomainServiceTests.swift
@@ -1,0 +1,83 @@
+//
+// This file is part of Canvas.
+// Copyright (C) 2025-present  Instructure, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Combine
+import Core
+import Foundation
+import Testing
+
+@testable import Horizon
+
+struct DomainServiceTests {
+
+    // Given the Domain Service
+    // When initilized with .cedar
+    // When initialized with the baseURL https://example.com
+    // When initialized with the us-east-1 region
+    // Then the url is https://cedar-api-production.us-east-1.temp.prod.inseng.io
+    @Test func when_initialized_with_east_1_region_then_url_is_correct() async throws {
+        // Given
+        let mockAPI = TestHooks.MockAPI()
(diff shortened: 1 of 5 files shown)


inst-danger:
<h3>Builds</h3><p><strong>Commit:</strong> refs: CLX-2815 affects: Student release note: Fixing a bug builds: student (<a href="https://github.com/instructure/canvas-ios/commit/73ebc92">73ebc92</a>)<br/><strong>Build Number:</strong> 196<br/><strong>Built At:</strong> Aug 29 21:06 CEST (08/29 01:06 PM MDT)</p><table><tr><td valign="top" width="250" align="center"><details open><summary> Student</summary><a href="https://app.bitrise.io/app/5a9c92bb-8a36-489b-85d9-8665096f5127/installable-artifacts/31b75cd3c31b79d8/public-install-page/01cae452b886adca1ca9bb3a0953686e"></a></details></td></tr></table>

Horizon/HorizonTests/Common/Data/API/DomainServiceTests.swift
```diff
@@ -0,0 +1,81 @@
...
+//
+
+import Combine
+import Core
+import Foundation
+import Testing
+
+@testable import Horizon
+
+struct DomainServiceTests {
```
szabinst: Please take a look at how other tests are structured. There's a base class called `CoreTestsCase` that provides all the common objects that one might want to mock. 

I like the use of the new `@Test` macro and `#expect` condition!

szabinst (changes requested):
The fix looks good! I think we don't need to mention this in the release notes; in cases like this we simply add `release notes: none`.

Let's follow the same folder naming pattern as other projects and use `Horizon/HorizonUnitTests`.

Horizon/HorizonTests/Common/Data/API/DomainServiceTests.swift
szabinst: Now that I'm looking into this new framework, I see that with structs, we can achieve namespacing. I'd be still interested to use the existing canvas environment while utilizing the framework to its best.

szabinst (dismissed):

szabinst:
I just noticed that there are occurrences of creating a LoginSession where we don't pass `canvasRegion`. Please double check if it's really not needed in those cases.

Horizon/HorizonTests/Common/Data/API/DomainServiceTests.swift
reabbotted: I will go ahead and change it. I do think that inheriting from the CoreTestCase is probably a bad practice though. It references a lot of things that most unit tests shouldn't care about, such as the Analytics handler and an in memory database. Referencing a data store like singleSharedTestDatabase really makes it more of an integration test. To keep the tests light and flexible they should only reference the mock dependencies that they need to, so it seems like it would be better to have builder functions for mock dependencies so the tests could pick and build only what's necessary.

reabbotted: It looks like the other targets, like Teacher and Parent inherit from their own base class, like StudentTestCase, that sets up an environment. Would you like me to do something similar for the Horizon target?

reabbotted: oh, nm, looks Ahmed already did it.

szabinst (approved):

PR #3628, opened 2025-09-08
[CLX-2833][Student] - Pendo Tagging Fix

reabbotted:
Fixing Pendo tagging for CLX

### DIFF
--- a/Student/Student/Info.plist
+++ b/Student/Student/Info.plist
@@ -66,6 +66,16 @@
 				<string>pendo-889d43f7</string>
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
+			<string>pendo-pairing-clx</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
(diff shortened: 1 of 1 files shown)


szabinst (commented):
I'm not sure if I understand why this change is needed for tagging? The Canvas (Student) app already has a bundle name set up in Pendo and I just tried launching the debugger, it launched successfully.

reabbotted:
> I'm not sure if I understand why this change is needed for tagging? The Canvas (Student) app already has a bundle name set up in Pendo and I just tried launching the debugger, it launched successfully.

for jenny and i it was not working. Make sure that you've deleted the old canvas horizon app on your phone. What was happening for us is if we selected the canvas student app on the Pendo site to start tagging, when we scanned the qr code, it wouldn't bring up the designer on the web. If we chose the CLX project in Pendo and tried to scan the qr code, it would tell us the scheme was invalid. I believe what it's doing is Pendo is expecting the Canvas Horizon bundle identifier (com.instructure.horizon) when we're on CLX pages. So we need to register the URL scheme for the Pendo CLX project so that when we scan the qr code, it opens the designer.

PR #3631, opened 2025-09-09
Hotfix - Todo

ndrsszsz:
## Hotfix - Todo

refs: MBL-19291
builds: All
affects: Student
release note: Fixed issues of Todo list.
test plan: See ticket.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Planner/Model/API/APIPlannable.swift
+++ b/Core/Core/Features/Planner/Model/API/APIPlannable.swift
@@ -32,6 +32,36 @@ public struct APIPlannable: Codable, Equatable {
     public let plannable_date: Date
     let submissions: TypeSafeCodable<APIPlannable.Submissions, Bool>?
     public let details: APIPlannable.Details?
+    public var plannableType: PlannableType { .init(rawValue: plannable_type) ?? .other }
+
+    public var context: Context? {
+        if let context = contextFromContextType() {
+            return context
+        }
+        if PlannableType(rawValue: plannable_type) == .planner_note {
+            // Notes have no 'context_type', but have IDs in the inner 'plannable' object
+            return contextFromInnerPlannableObject()
+        }
+        return nil
+    }
+
+    private func contextFromContextType() -> Context? {
+        guard let raw = context_type, let type = ContextType(rawValue: raw.lowercased()) else {
+            return nil
+        }
+        return switch type {
+        case .course: Context(.course, id: course_id?.rawValue)
+        case .group: Context(.group, id: group_id?.rawValue)
+        case .user: Context(.user, id: user_id?.rawValue)
+        default: nil
+        }
+    }
+
+    private func contextFromInnerPlannableObject() -> Context? {
+        // order matters: 'course_id' has precedence over 'user_id'
+        return Context(.course, id: self.plannable?.course_id)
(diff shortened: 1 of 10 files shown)


Core/Core/Features/Planner/Model/API/APIPlannable.swift
```diff
@@ -205,23 +236,23 @@ extension APIPlannerOverride {
...
+public var path: String {
+    if let userID {
+        return "users/\(userID)/planner/items"
+    } else {
+        return "planner/items"
     }
+}
 
-    public var query: [APIQueryItem] {
+public var query: [APIQueryItem] {
```
vargaat: Indentation is off.

ndrsszsz: Thanks. Fixed now.

vargaat (approved):
QA+1

Core/Core/Features/Planner/Model/CoreData/Plannable.swift
```diff
@@ -38,6 +38,8 @@ public final class Plannable: NSManagedObject {
     @NSManaged public var pointsPossibleRaw: NSNumber?
     @NSManaged public var userID: String?
     @NSManaged public var details: String?
+    @NSManaged public var isMarkedComplete: Bool
+    @NSManaged public var isSubmitted: Bool
```
rh12, line 42: These are optional in the CoreData entity, but not here.
Plus no values are set in `save(_ item: APIPlannerNote...)` and `save(_ item: APICalendarEvent...)`.
IMHO these should not be optional in CoreData either (with probably default: false), and they should have assigned values in all `save()` methods.

Core/Core/Features/Todos/Model/TodoInteractor.swift
```diff
@@ -37,7 +37,7 @@ public final class TodoInteractorLive: TodoInteractor {
 
     private var subscriptions = Set<AnyCancellable>()
 
-    init(startDate: Date = .now, endDate: Date = .distantFuture, env: AppEnvironment) {
+    init(startDate: Date = .now, endDate: Date = .now.addDays(28), env: AppEnvironment) {
```
rh12: 1. The date range does not match the one in the widget.
There it's now.startofDay and +28 days to THAT. so we deal with whole days there, which is better IMHO.

2. We should use `Clock.now` to support unit testing.

Core/Core/Features/Planner/Model/CoreData/Plannable.swift
```diff
@@ -58,20 +60,29 @@ public final class Plannable: NSManagedObject {
         set { canvasContextIDRaw = newValue?.canvasContextID }
     }
 
+    public var shouldShowInTodoList: Bool {
+        plannableType != .announcement && plannableType != .assessment_request && !isMarkedComplete && !isSubmitted
```
rh12, line 65: Please unit test this logic either in this PR or the other one which is currently open.
IMHO we should also mention  the 28 days window in a comment here, and that it is enforced elsewhere.
(Since we have a property which claims to be the source of truth for deciding whether to show this event or not in ToDos)

PR #3639, opened 2025-09-12
[MBL-19292] Update Nutrient SDK

petkybenedek:
Upgrade Nutrient (PSPDFKit) to version 14.2. Version 26.0 is the newest, but is incompatible with Xcode 16.4. We should upgrade to the newest version when we upgrade to Xcode 26.

Checked all annotation features and they work as expected.

refs: MBL-19292
builds: Student, Parent, Teacher
affects: Student, Parent, Teacher
release note: none

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [x] Tested on tablet
- [x] Tested in dark mode
- [x] Tested in light mode
- [x] Approve from product

### DIFF
--- a/Canvas.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Canvas.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "257ce0b596a02f742731a202b60715b4aacdb586de9dd5e4871c9baa633e1e71",
+  "originHash" : "102d90cdfa1716b86d908a5b9c78094cc7e8ac6573d7ecb6e8ea7a20d67852e4",
   "pins" : [
     {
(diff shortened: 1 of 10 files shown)


petkybenedek:
The Bitrise builds doesn't have Nutrient license installed, that's why the toolbar is hidden in this case. If you build it locally with a license, it works correctly so we can ignore the mentioned problem.

balintbartok:
I'm not sure about the bitrise toolbar issue, but it's not visible for other PRS either, so I'm convinced it's true

QA + 1

balintbartok (approved):

suhaibabsi-inst (approved):
QA +1

Tested locally with no major issues.
Apparently, the Bitrise issue exists even before this PR, as shown in "ad-hoc builds" workflow logs.

We will need to revisit it later as we upgrade to version 26.

PR #3648, opened 2025-09-17
[MBL-19317][Teacher] SpeedGrader - Submissions List Score-Based Filtering 

suhaibabsi-inst:
refs: MBL-19317
affects: Teacher
builds: Teacher
release note: Adds filtering options by status, section & score to submissions list screen.

## Important notes

- Hiding "Filter by Section" when there is only one section defined for the course viewed. Because it is required for any user of the course to be enrolled to at least one section. 

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


inst-danger:
<h3>Builds</h3><p><strong>Commit:</strong> Basic functions of score-based filtering (<a href="https://github.com/instructure/canvas-ios/commit/16049ca">16049ca</a>)<br/><strong>Build Number:</strong> 365<br/><strong>Built At:</strong> Sep 17 14:35 CEST (09/17 06:35 AM MDT)</p><table><tr><td valign="top" width="250" align="center"><details open><summary> Teacher</summary><a href="https://app.bitrise.io/app/5a9c92bb-8a36-489b-85d9-8665096f5127/installable-artifacts/f91c167b75bc3358/public-install-page/5f082aebbdd439d328e4addb9dab407e"></a></details></td></tr></table>

Core/Core/Common/Extensions/Combine/SubjectExtensions.swift
```diff
@@ -0,0 +1,29 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Combine
+import SwiftUI
+
+public extension CurrentValueSubject {
```
rh12: Shouldn't this be restricted to `Never`?
This way it's very easy to silently forget about errorhandling

Core/Core/Common/Extensions/Foundation/StringExtensions.swift
```diff
@@ -190,6 +190,17 @@ extension String {
     public var doubleValue: Double? {
         StringFormatUtils.numberFormatter.number(from: self)?.doubleValue
     }
+
+    public var urlQuerySafePercentEncoded: String? {
+        addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
+    }
+
+    public func asURLPathWithQuery(_ queryItems: [URLQueryItem]) -> String {
+        var comps = URLComponents()
```
rh12:
```suggestion
        var components = URLComponents()
```

Teacher/Teacher/Submissions/SubmissionsList/Model/SubmissionListInteractor.swift
```diff
@@ -21,19 +21,31 @@ import Core
...
     var course: AnyPublisher<Course?, Never> { get }
+    var courseSections: AnyPublisher<[CourseSection], Never> { get }
     var assigneeGroups: AnyPublisher<[AssigneeGroup], Never> { get }
 
     var context: Context { get }
     var assignmentID: String { get }
 
     func refresh() -> AnyPublisher<Void, Never>
-    func applyFilter(_ filter: GetSubmissions.Filter)
+    func applyPreference(_ pref: SubmissionListPreference)
```
rh12:
```suggestion
    func applyPreferences(_ pref: SubmissionListPreferences)
```

Teacher/Teacher/Submissions/SubmissionsList/View/SubmissionsFilterScreen.swift
```diff
@@ -19,83 +19,90 @@
...
+                MultiSelectionView(
+                    title: String(localized: "Statuses", bundle: .teacher),
+                    identifierGroup: "SubmissionsFilter.statusOptions",
+                    options: viewModel.statusFilterOptions
+                )
+                .tint(viewModel.courseColor)
+
+                if let gradeInputType = viewModel.gradeInputType {
+
+                    Section {
```
rh12, line 43: Please use `InstUI.ListSectionHeader` instead

Teacher/Teacher/Submissions/SubmissionsList/View/SubmissionsFilterScreen.swift
```diff
@@ -19,83 +19,90 @@
...
-            .tint(courseColor)
-            Spacer()
+        ScrollView {
+            VStack(alignment: .leading, spacing: 0) {
+                MultiSelectionView(
+                    title: String(localized: "Statuses", bundle: .teacher),
+                    identifierGroup: "SubmissionsFilter.statusOptions",
+                    options: viewModel.statusFilterOptions
+                )
+                .tint(viewModel.courseColor)
```
rh12: This could be moved out one level, as it applies to everything.

Teacher/Teacher/SpeedGrader/Common/GradeInputTextFieldCell.swift
```diff
@@ -153,12 +153,12 @@ extension GradeInputTextFieldCell {
         isSaving: CurrentValueSubject<Bool, Never>
     ) {
         let subtitle: String? = switch inputType {
-        case .points: nil
-        case .percentage: "(\(pointsPossibleText))"
+        case .percentage where pointsPossibleText.isNotEmpty: "(\(pointsPossibleText))"
+        default: nil
```
rh12: Should we even display these rows when pointspossible is empty?
Is it even possible? We always deal in scores, even for other grading types, do we not? summoning @vargaat 

If we still keep this please replace `default` with `.percentage`

Same things for a11y below

Teacher/Teacher/Submissions/SubmissionsList/ViewModel/SubmissionsFilterViewModel.swift
```diff
@@ -0,0 +1,177 @@
...
+            initialId: listViewModel.sortMode.rawValue
+        )
+
+        gradeInputType = listViewModel.gradeInputType
+
+        scoredMoreFilterValue = .init(listViewModel
+            .scoreBasedFilters
+            .moreThanFilter
+            .flatMap({ listViewModel.toScoreInputValue($0.score) })?
+            .formatted(.number.precision(.fractionLength(0 ... 3))) ?? ""
```
rh12: IMHO we should use the exact same formatting we use in SpeedGrader.
Could we please use GradeFormatter.numberFormatter/percentFormatter instead?
For example this provides 3 digits precision, while we use 2 digits there.
The formatter implementation may need to be replaced later on, but introducing more implementations for the same instead of centralizing it is something we should avoid.

Teacher/Teacher/SpeedGrader/Common/GradeInputTextFieldCell.swift
vargaat: In the end all assignments are translated to points. However, when the max point is 0 we could hide the score based filtering, but I think this is a nice to have feature.

suhaibabsi-inst: This is done mainly to match Figma design. So, perhaps Marcell would have a say on it

Teacher/Teacher/Submissions/SubmissionsList/ViewModel/SubmissionsFilterViewModel.swift
suhaibabsi-inst: Will be using `GradeFormatter.numberFormatter`.
As it is mainly used translate numeric value to text input form of a text. Suffixes "%" or "points" is added by the cell view logic.

Teacher/Teacher/SpeedGrader/Common/GradeInputTextFieldCell.swift
rh12: On Figma the label always says "Score" and the textfield always expects points.
We should always expect a "Score" here (which always means points, as opposed to "Grade" which can be many things)
On a related note, I see we don't display these textfields for LetterGrades, but we should.
Regarding `points_possible: 0`, I think it should be simply displayed as well. It's an edge case anyway and it makes sense to show it to the teacher in case the Teacher choses to set points possible to zero, instead of making the assignment not gradable. (for whatever reason)
So all in all, we should always display these textfields if the assignment is gradable (`points_possible` is not `null`)

Teacher/Teacher/Submissions/SubmissionsList/View/SubmissionsFilterScreen.swift
```diff
@@ -19,83 +19,87 @@
...
+                    options: viewModel.statusFilterOptions
+                )
+
+                if let gradeInputType = viewModel.gradeInputType {
+
+                    Section {
+                        VStack(alignment: .leading, spacing: 0) {
+                            GradeInputTextFieldCell(
+                                title: String(localized: "Scored More than", bundle: .teacher),
+                                inputType: gradeInputType,
```
rh12: 1. This should always be points, as discussed in other comment
2. Please make sure its always interpreted as score (not as percentage for example)
3. Same for the other cell below
4. Both of them should be visible for all grading types (when pointpossible is not nil)

rh12 (dismissed):
QA + 1, thanks for the changes!

rh12 (approved):

PR #3663, opened 2025-09-24
[MBL-19121] [S/P] DCP - Update Grade List

rh12:
refs: [MBL-19121](https://instructure.atlassian.net/browse/MBL-19121)
affects: Student, Parent
release note: Discussion checkpoints are now handled in the grade list.

- Updated Grade List to support DCPs
  - Reused views from Assignment List
  - Spent some minimal effort to keep supporting the already unused and not working WhatIf implementation
- Fixed sections ordering in Grade List (now it matches ordering on web, just like in Assignment List)

## Screenshots
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-19121]: https://instructure.atlassian.net/browse/MBL-19121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Common/CommonUI/InstUI/Views/CollapsibleListSection.swift
+++ b/Core/Core/Common/CommonUI/InstUI/Views/CollapsibleListSection.swift
@@ -33,10 +33,12 @@ extension InstUI {
 
         private let headerAccessibilityLabel: String
         private let listLevelAccessibilityLabel: String?
+        private let headerIdentifier: String?
 
         public init(
             title: String,
             customAccessibilityLabel: String? = nil,
+            headerIdentifier: String? = nil,
             itemCount: Int?,
             paddingSet: InstUI.Styles.PaddingSet = .sectionHeader,
             accessoryIconSize: CGFloat = 18,
@@ -46,6 +48,7 @@ extension InstUI {
             self.init(
                 label: Text(title),
                 accessibilityLabel: customAccessibilityLabel ?? title,
+                headerIdentifier: headerIdentifier,
                 itemCount: itemCount,
                 paddingSet: paddingSet,
                 accessoryIconSize: accessoryIconSize,
@@ -57,6 +60,7 @@ extension InstUI {
         public init(
             label: Label,
             accessibilityLabel: String,
+            headerIdentifier: String? = nil,
             itemCount: Int?,
             paddingSet: InstUI.Styles.PaddingSet = .sectionHeader,
             accessoryIconSize: CGFloat = 18,
@@ -77,6 +81,8 @@ extension InstUI {
             ].accessibilityJoined()
 
             self.listLevelAccessibilityLabel = itemCount.map(String.format(accessibilityListCount:))
+
+            self.headerIdentifier = headerIdentifier
         }
 
         // MARK: - Body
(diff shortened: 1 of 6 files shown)


Core/Core/Features/Assignments/AssignmentList/Model/StudentAssignmentListItem.swift
```diff
@@ -40,15 +41,24 @@ struct StudentAssignmentListItem: Equatable, Identifiable {
...
         assignment: Assignment,
+        userId: String?,
         dueDateTextsProvider: AssignmentDueDateTextsProvider = .live
     ) {
+        let submission: Submission?
+        if let userId {
+            submission = assignment.submissions?.first { $0.userID == userId }
+        } else {
+            submission = assignment.submission
+        }
```
suhaibabsi-inst, line 60:
```suggestion
        let submission: Submission? = if let userId {
            assignment.submissions?.first { $0.userID == userId }
        } else {
            assignment.submission
        }
```

suhaibabsi-inst (commented):
Code +1

Core/Core/Features/Assignments/AssignmentList/Model/StudentAssignmentListItem.swift
rh12: I think we discussed this somewhere already (with @vargaat definitely, but probably on some PR), that these assignments are kind of weird, so we prefer not to use this format when assigning value to a variable/property.
Either the "old" way is preferred or putting the whole thing in a `{ if ... }()` block.
(Same thing for switches)
This could probably go in the styleguide sometime.
So I would leave this as it is.

suhaibabsi-inst (approved):
Tested on iPhone XS, iOS 18.3.1.

QA +1 for general UI features, though since there still API work to be done later on grading, I only managed to check state of no grades is shown for them.

rh12:
All E2E tests passed. The failing one in student is not related (the DCP seeding still doesn't work)

Core/Core/Features/Assignments/AssignmentList/Model/AssignmentListSection.swift
```diff
@@ -27,11 +27,13 @@ struct AssignmentListSection: Equatable, Identifiable {
     enum Row: Equatable, Identifiable {
         case student(StudentAssignmentListItem)
         case teacher(TeacherAssignmentListItem)
+        case gradeListRow(StudentAssignmentListItem)
 
         var id: String {
             switch self {
             case .student(let row): row.id
             case .teacher(let row): row.id
+            case .gradeListRow(let row): row.id
```
vargaat, line 36: Let's add some explanatory comments what each of those are used for because it's not straightforward.

vargaat (approved):
QA+1

Core/Core/Features/Assignments/AssignmentList/Model/AssignmentListSection.swift
rh12: Good point! As discussed I will add these in my upcoming PR, because multiple branches are based on this one.

PR #3672, opened 2025-09-30
[MBL-19362] [S/T] Resolve Issue of Inbox Message Unblocked for Concluded Courses

suhaibabsi-inst:
refs: MBL-19362
affects: Student, Teacher
builds: Student, Teacher
release note: Fixed issue where message sending is not being disabled in UI for concluded courses

Basically the resolution to this, as agreed on Slack conversation, is to mimic the exact behavior on Web, which is to show course showing a SnackBar message when tapped indicating blocked selection.

## Test Plan

See [ticket's](https://instructure.atlassian.net/browse/MBL-19362) description

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

### DIFF
--- a/Core/Core/Features/Inbox/InboxCoursePicker/View/InboxCoursePickerView.swift
+++ b/Core/Core/Features/Inbox/InboxCoursePicker/View/InboxCoursePickerView.swift
@@ -20,7 +20,6 @@ import SwiftUI
 
 public struct InboxCoursePickerView: View {
     @ObservedObject private var viewModel: InboxCoursePickerViewModel
-    @StateObject var snackBarViewModel = SnackBarViewModel()
 
     @Environment(\.dynamicTypeSize) private var dynamicTypeSize
 
@@ -46,7 +45,7 @@ public struct InboxCoursePickerView: View {
         .frame(maxWidth: .infinity)
         .navigationBarStyle(.modal)
         .background(Color.backgroundLightest)
-        .snackBar(viewModel: snackBarViewModel)
+        .snackBar(viewModel: viewModel.snackbarViewModel)
     }
 
     @ViewBuilder
@@ -125,14 +124,7 @@ public struct InboxCoursePickerView: View {
         let accessibilityLabel = isSelected(course) ? Text("Selected: \(courseName)", bundle: .core) : Text(courseName)
         return VStack(spacing: 0) {
             Button {
-                viewModel.onSelect(selected: course, onInvalidated: {
-                    snackBarViewModel.showSnack(
-                        String(
-                            localized: "Course concluded. Unable to send messages!",
-                            bundle: .core
-                        )
-                    )
-                })
+                viewModel.onSelect(selected: course)
             } label: {
                 HStack {
                     Circle()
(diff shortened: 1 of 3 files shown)


Core/Core/Features/Inbox/InboxCoursePicker/ViewModel/InboxCoursePickerViewModel.swift
```diff
@@ -49,9 +49,14 @@ class InboxCoursePickerViewModel: ObservableObject {
         setupOutputBindings()
     }
 
-    public func onSelect(selected: Course) {
-        let context = RecipientContext(course: selected)
-        onSelect(selected: context)
+    public func onSelect(selected: Course, onInvalidated: () -> Void = {}) {
+        if selected.isPastEnrollment {
```
vargaat, line 54: `isPastEnrollment` has this comment: 

> Use with caution! This property doesn't take section dates or the actual enrollment's concluded state into account.

Can we safely use it here?

suhaibabsi-inst: Yes, this is sufficient to exactly mimic the behavior on the Web version. 
In other words, on Web, user is not being blocked if he is enrolled to a course section that ends in the past.

suhaibabsi-inst: One note through, course get already excluded from that picker menu when user is enrolled in a section with a past end date, with `Students can only participate in the course between these dates` checked. On both Web & Mobile.

Core/Core/Features/Inbox/InboxCoursePicker/View/InboxCoursePickerView.swift
```diff
@@ -142,6 +152,7 @@ public struct InboxCoursePickerView: View {
 
                 }
                 .foregroundStyle(Color.textDarkest)
+                .opacity(course.isPastEnrollment ? 0.6 : 1)
```
rh12: We have a `.disabledGray` color, please use that one instead of opacity change.

Core/Core/Features/Inbox/InboxCoursePicker/View/InboxCoursePickerView.swift
```diff
@@ -122,7 +125,14 @@ public struct InboxCoursePickerView: View {
         let accessibilityLabel = isSelected(course) ? Text("Selected: \(courseName)", bundle: .core) : Text(courseName)
         return VStack(spacing: 0) {
             Button {
-                viewModel.onSelect(selected: course)
+                viewModel.onSelect(selected: course, onInvalidated: {
```
rh12: IMHO this `onInvalidated` within the `onSelect` is kind of confusing.
WDYT about passing the viewController as a parameter and handle the showing from the viewModel?
You can search for usages of `.showSnack()` for examples.

suhaibabsi-inst: Since, the use of `snackBar(viewModel:)` is common place, I will be passing the `snackBarModel` instead of the viewController. As `CoreHostingController` is not compliant to `SnackBarProvider` by default.
Or, perhaps, an instance for the view itself.

suhaibabsi-inst: After few experiments, 
I think we better keep this signature as is, as it lends itself more easily for unit testing.

suhaibabsi-inst: And this way we keep UI related things in the view and away from the viewModel.

vargaat: I moved the snack view model into the picker view model and let the picker view model control the snack message. This way the snack message itself became testable.

vargaat (approved):
QA+1

PR #3686, opened 2025-10-07
[MBL-18172] [S/T] Split View selection highlights

petkybenedek:
### What's new?
Selections stay highlighted in Split View on the following screens:
- Assignments
- Grades
- Quizzes
- To-Do
- Notifications
- Inbox

This PR introduces a SwiftUI-idiomatic approach to indicate selection with Environment Values. This eliminates the need of passing down parameters in the View Hierarchy.

refs: [MBL-18172](https://instructure.atlassian.net/browse/MBL-18172)
builds: Student, Teacher
affects: Student, Teacher
release note: Selections on iPad now stay highlighted.

## Screenshots

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


Core/Core/Features/Assignments/AssignmentList/View/AssignmentListView.swift
```diff
@@ -19,24 +19,25 @@
...
     private let whatIfModel: GradeListWhatIfModel?
+	private let selectedAssignmentID: String?
 
     init(
         sections: [AssignmentListSection],
         identifierGroup: String,
-        navigateToDetailsAction: @escaping (URL?) -> Void,
-        whatIfModel: GradeListWhatIfModel? = nil
+        navigateToDetailsAction: @escaping (URL?, String) -> Void,
+        whatIfModel: GradeListWhatIfModel? = nil, selectedAssignmentID: String? = nil
```
suhaibabsi-inst: How about replacing the whole `navigateToDetailsAction` & `selectedAssignmentID` with a `@Published` property `selectedAssignmentID` on screen viewModel. and you can pass that property to this view as `@Binding`.  Having an action and a property referring to the same thing seems difficult to maintain. And you get even more SwiftUI-friendly way of handling state.

suhaibabsi-inst: We can say the same about other list views as well: `InboxView` and `GradeListScreen`

Core/Core/Common/CommonUI/ListSelection/Selection.swift
```diff
@@ -0,0 +1,46 @@
...
+		modifier(SelectionModifier(isSelected: condition))
+	}
+
+	/// This View Modifier sets the `isSelected` Environment Value on a view to `false`.
+	func selectionIndicatorDisabled() -> some View {
+		self.environment(\.isItemSelected, false)
+	}
+}
+
+struct SelectionModifier: ViewModifier {
```
rh12: Could you please
1. Move this whole file into ViewModifiers (as ItemSelectionModifier)
2. Make the `@Entry public var` property public
3. Make the modifier methods public
4. Make the modifier struct private

Core/Core/Common/CommonUI/ListSelection/Selection.swift
```diff
@@ -0,0 +1,46 @@
...
+	@Environment(\.viewController) private var controller
+	let isSelected: Bool
+
+	func body(content: Content) -> some View {
+		content
+			.environment(\.isItemSelected, controller.value.splitViewController?.isCollapsed == false && isSelected)
```
rh12:
```suggestion
        let isSplitViewControllerCollapsed = controller.value.splitViewController?.isCollapsed ?? true
		content
			.environment(\.isItemSelected, isSplitViewControllerCollapsed && isSelected)
```
IMHO `?? true` is much more clear than `== false`

Core/Core/Features/Assignments/AssignmentList/View/AssignmentListView.swift
rh12: 1. add parameter in new line, and maybe before the navigate action (unless you change these into bindings)
2. use `somethingId` instead of `somethingID`, elsewhere too

Core/Core/Features/Assignments/AssignmentList/View/AssignmentListView.swift
```diff
@@ -57,29 +58,34 @@ struct AssignmentListView: View {
...
     @ViewBuilder
     private func studentCell(model: StudentAssignmentListItem, isLastItem: Bool) -> some View {
-        let routeAction = { navigateToDetailsAction(model.route) }
+		let routeAction = { navigateToDetailsAction(model.route, model.id) }
         let itemIdentifier = "\(itemIdentifierGroup).\(model.id)"
 
         if let subItems = model.subItems {
             InstUI.CollapsibleListRow(
                 cell: StudentAssignmentListItemCell(model: model, isLastItem: nil, action: routeAction)
+					.selected(when: model.id == selectedAssignmentID)
```
rh12: This doesn't seem to be needed. the environment value is already set one level above.
Same for teacher

Core/Core/Features/Assignments/AssignmentList/View/AssignmentListView.swift
```diff
@@ -57,29 +58,34 @@ struct AssignmentListView: View {
...
                     .identifier(itemIdentifier),
                 isInitiallyExpanded: false
             ) {
                 ForEach(subItems) { subItem in
                     StudentAssignmentListSubItemCell(model: subItem, action: routeAction)
                         .identifier(itemIdentifier, subItem.tag)
+						.selectionIndicatorDisabled()
```
rh12: please move this before the `.identifier`, for consistency

Core/Core/Features/Assignments/AssignmentList/View/AssignmentListView.swift
petkybenedek: On the AssignmentListScreen, the navigation action is still required, that's how the ViewModel receives the viewController. Without it, your approach would work, but now I have to go with my solution.

rh12 (approved):
QA + 1

As discussed offline, we scoped the changes to SwiftUI screens only, so not all screens in the PR description are updated.

One minor issue:
On iPhone, Inbox rows are not showing highlight on tap

suhaibabsi-inst (approved):
QA +1

Tested using iPad simulator on iOS 18.5.

I was able to see highlighted rows as selected for most inner pages of course details (Assignments, Grades, Pages, Modules, .. etc) except for Quizzes which is a non-SwiftUI list.

Also, screens Inbox and Notifications was working as expected as well.

For To-DO screen, it is still not working for being non-SwiftUI list.

Approving since it is already stated that non-SwiftUI lists are not covered in this PR.

PR #3927, opened 2026-03-05
[MBL-19819][S] Offline Mode - Enhanced Course Sync failure reporting for embedded media content

suhaibabsi-inst:
refs: MBL-19819
affects: Student
builds: Student
release note: Enhanced course Offline sync failure reporting for embedded media content.

- **Distinguishes embedded media failures from full sync failures.** Previously, when embedded media (images/files referenced in HTML content, Studio media) failed to download due to 403/404 errors, they were silently swallowed. Now these failures are tracked and surfaced distinctly from hard sync errors.
- **New `embeddedContentErrorCourseIds` field** added to `CDCourseSyncDownloadProgress` (Core Data) to persist which courses had partial embedded content failures.
- **`HTMLParser` emits failures via `embeddedContentFailurePublisher`** instead of silently replacing errors with empty strings. 403/404 errors on embedded assets are still non-fatal (download continues), but the affected course IDs are now recorded. Also, when locked files are had a link in one of the pages, they won't be downloaded, and they'd also be considered as secondary content failure.
- **Studio media interactor** similarly propagates partial failure course IDs upstream.
- **UI differentiation:** Courses with embedded content failures show a warning state (`.textWarning` color) instead of an error state, and the sync progress widget shows an "Incomplete" warning banner (yellow) instead of a failure banner (red).
- **Notification differentiation:** A new `sendOfflineSyncEmbeddedContentWarningNotification` local notification is sent when only embedded content failed, rather than the full failure notification.

## Screenshots

<table>
<tr><th>1</th><th>2</th></tr>
<tr>
<td></td>
<td></td>
</tr>
<tr><th>3</th><th>4</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Test Plan

- [x] **Full sync success:** Sync a course with no inaccessible media → progress screen shows all green, success notification fires, no warning state.
- [x] **Embedded media failure:** Sync a course where some embedded HTML images or files return 403/404 → course row shows warning indicator (yellow/warning color, not red error), sync completes overall, and the "Offline Content Sync Incomplete" warning notification fires.
- [x] **Full sync failure:** Trigger a hard sync failure (e.g. network drop mid-sync) → course row shows red error state and "Offline Content Sync Failed" notification fires as before.
- [x] **Studio media failure:** Sync a course with Studio media that is inaccessible → course appears in embedded content failure list (warning, not hard error).
- [ ] **OS interrupt:** Force the app to background mid-sync → "Offline sync was interrupted" error notification fires and no embedded content warning is shown.
- [x] **Progress widget:** After an embedded media failure sync, the dashboard widget shows yellow "Offline Content Sync Incomplete" banner with the correct subtitle text.
- [x] **Persistence:** Kill and relaunch the app after an embedded media failure sync → the warning state persists correctly from Core Data.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product


Core/Core/Features/CourseSync/Common/View/OfflineListCellView.swift
```diff
@@ -32,10 +32,14 @@ public struct OfflineListCellView: View {
...
     enum State {
+        static func error(_ message: String?) -> Self {
+            .error(message, false)
+        }
+
         case idle
         case loading(Float?)
         case downloaded
-        case error(String?)
+        case error(String?, Bool)
```
vargaat: Let's add a comment on what this Bool is. I see on line 172 that this is treated as a warning flag but I have to look up its usage to find out the reason for this.

Core/Core/Features/CourseSync/CourseSyncDownloader/Model/Downloaders/CourseSyncStudioMediaInteractor.swift
```diff
@@ -66,32 +67,33 @@ public class CourseSyncStudioMediaInteractorLive: CourseSyncStudioMediaInteracto
...
-                .map { mediaData }
+                    return cleanupInteractor
+                        .removeNoLongerNeededVideos(
+                            allMediaItemsOnAPI: Array(mediaItems),
+                            mediaLTIIDsUsedInOfflineMode: Array(idsToDownload),
+                            offlineStudioDirectory: studioDirectory
+                        )
+                }
+                .collect()
+                .flatMap({ _ in allCoursesMedia.publisher })
```
vargaat: Let's be consistent on the parenthesis and omit them when they can be omitted, just like you did on line 79.

Core/Core/Features/CourseSync/CourseSyncDownloader/Model/Downloaders/CourseSyncStudioMediaInteractor.swift
```diff
@@ -107,22 +109,37 @@ public class CourseSyncStudioMediaInteractorLive: CourseSyncStudioMediaInteracto
...
+                            name: "Studio Offline Sync Failed for Course",
+                            reason: error.debugDescription
+                        )
+
+                        return Publishers.typedJust(mediaData.courseID)
+                    }
+                    .eraseToAnyPublisher()
             }
-            .catch { error -> AnyPublisher<Void, Never> in
+            .compactMap({ $0 })
```
vargaat:
```suggestion
            .compactMap { $0 }
```

Core/Core/Features/CourseSync/CourseSyncDownloader/Model/Downloaders/CourseSyncStudioMediaInteractor.swift
```diff
@@ -107,22 +109,37 @@ public class CourseSyncStudioMediaInteractorLive: CourseSyncStudioMediaInteracto
...
-            .flatMap { [weak self] mediaData in
-                guard let self else { return Publishers.noInstanceFailure(output: Void.self) }
+            .flatMap(removeNoLongerNeededMedia)
+            .flatMap { [weak self] mediaData -> AnyPublisher<CourseSyncID?, Error>  in
+                guard let self else {
+                    return Publishers.noInstanceFailure(output: CourseSyncID?.self)
+                }
+
                 return self.downloadMediaTweakingIFrameReferences(mediaData)
+                    .map({ nil as CourseSyncID? })
```
vargaat:
```suggestion
                    .map { nil as CourseSyncID? }
```

Core/Core/Features/CourseSync/CourseSyncDownloader/Model/HTMLParser/HTMLParser.swift
```diff
@@ -60,6 +66,7 @@ public class HTMLParserLive: HTMLParser {
             .eraseToAnyPublisher()
     }
 
+    // swiftlint:disable:next function_body_length
```
vargaat: Can't we extract the newly added error handling logic into a common place to save up some lines? Those two error handlers seem to be copy-pasted, so I guess it would be easy to refactor.

Core/Core/Features/CourseSync/CourseSyncSelector/Model/Entities/CourseSyncSelectorCourse.swift
```diff
@@ -62,4 +62,23 @@ public final class CourseSyncSelectorCourse: NSManagedObject {
...
+            Context.course(courseId).canvasContextID
+        )
+
+        let count = context.count(of: Tab.self, predicate: predicate)
+
+        guard tabs.count != count else { return }
+
+        let tabs: [Tab] = context.fetch(predicate)
+        self.tabs = Set(tabs)
+    }
```
vargaat: What's the root cause of this? Shouldn't we update this relationship in `Tab.save()`?

Core/Core/Features/CourseSync/Common/Model/Entities/CDCourseSyncStateProgress.swift
```diff
@@ -61,6 +61,7 @@ final class CDCourseSyncStateProgress: NSManagedObject, Comparable {
             case 0: return .loading(progress?.floatValue)
             case 1: return .error
             case 2: return .downloaded
+            case 3: return .downloaded(isEmbeddedMediaComplete: false)
```
vargaat: What's the difference between `.downloaded` and `.downloaded(isEmbeddedMediaComplete: true)`? They seem to be redundant. Can't we just keep the one with `isEmbeddedMediaComplete`?

Core/Core/Features/CourseSync/CourseSyncSelector/Model/Entities/CourseSyncSelectorCourse.swift
suhaibabsi-inst: The broken relationship can happen when Tabs are saved after API request in `GetContextTabs` use case.
We have only this type having such relationship to course `[Tab]`s, so we either need to fix this relationship here, or in `GetContextTabs.write()` method. 

This broken relationship issue could show up when Course Sync Progress screen is shown through tapping on sync issue dashboard card.

vargaat: I'd suggest to continue our practice and update the relationship in `Tab.save` if possible.

Core/Core/Common/CommonModels/Contexts/Tab.swift
```diff
@@ -70,5 +70,23 @@ public class Tab: NSManagedObject {
...
+            let courseId = context.courseId,
+            let course: CourseSyncSelectorCourse = client
+            .first(
+                scope: .where(
+                    #keyPath(CourseSyncSelectorCourse.courseId),
+                    equals: courseId
+                )
+            )
+        else { return }
+        course.tabs.insert(self)
```
vargaat: We have the relationship defined in the database for the `Tab` entity under the name `course`. We could add this property to the actual Tab file so the connection would be more visible. Setting the relationship here would reduce to `self.course = course`.

Core/Core/Common/CommonModels/Contexts/Tab.swift
```diff
@@ -59,7 +59,7 @@ public class Tab: NSManagedObject {
         TabName(rawValue: id) ?? .custom
     }
 
-    func save(_ item: APITab, in client: NSManagedObjectContext, context: Context) {
+    func save(_ item: APITab, in client: NSManagedObjectContext, context: Context, fixingCourseSyncRelationship: Bool = true) {
```
vargaat: Why the conditional fixing? This is a relationship that should be valid and up-to-date all the time.

suhaibabsi-inst: Yes, this is to avoid collision with the setting `tabs` property in `CourseSyncSelectorCourse.save()`.

suhaibabsi-inst: This is passed `fixingCourseSyncRelationship=false` only in that method of `CourseSyncSelectorCourse`

rh12 (dismissed):
Code + 1

rh12 (commented):
Code + 1

vargaat (approved):
QA+1

rh12 (approved):

PR #3933, opened 2026-03-11
[S] Fix manual login

Ahmed-Naguib93:
[ignore-commit-lint]
builds: Student
affects: Student
release note: none

test plan: none

### DIFF
--- a/Core/Core/Features/Login/MobileVerify.swift
+++ b/Core/Core/Features/Login/MobileVerify.swift
@@ -19,7 +19,11 @@
 import Foundation
 
 enum MobileVerify {
-    static var strategy: MobileVerifyStrategy = DefaultMobileVerifyStrategy()
+    static func strategy(for host: String) -> MobileVerifyStrategy {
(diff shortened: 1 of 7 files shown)


Core/Core/Features/Login/MobileVerify.swift
```diff
@@ -19,7 +19,11 @@
 import Foundation
 
 enum MobileVerify {
-    static var strategy: MobileVerifyStrategy = DefaultMobileVerifyStrategy()
+    static func strategy(for host: String) -> MobileVerifyStrategy {
+        host.contains("horizon")
```
vargaat: There are a number of institutions with `horizon` in their host name so this is not a good check.

vargaat (changes requested):

Core/Core/Features/Login/MobileVerify.swift
Ahmed-Naguib93: Changed

PR #3975, opened 2026-04-04
[CLXR-462][S] Add career offline sync

Ahmed-Naguib93:
feat: add offline sync
refs: CLXR-462
builds: student
affects: student
release note: none

1.  Adds Horizon-specific background offline sync infrastructure to the Canvas Career (Horizon) experience                                                                                                         
2. Introduces HSyncBackgroundTask that iterates logged-in sessions, refreshes course data in the background, and restores the previously active session on completion                                             
3. Adds HSyncNextDateInteractor and HSyncAccountsInteractor to calculate which sessions are due for sync and when the next background task should fire                                                            
4. Adds HSyncScheduleInteractor to schedule/reschedule the BGProcessingTask after each sync cycle                                                                                                                 
5.  Stores Horizon sync preferences (frequency, nextDate, autoSyncEnabled, offlineSyncItems) in SessionDefaults per user session                                                                                   
6. Registers the background task identifier (com.instructure.icanvas.horizon-sync) in Info.plist and StudentAppDelegate                                                                                           
7.  Routes background sync scheduling in applicationDidEnterBackground based on whether the active app is Horizon or Canvas Academic


Horizon/Horizon/Sources/Features/OfflineSync/BackgroundTask/HSyncBackgroundTaskRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import BackgroundTasks
+import Core
+
+public final class HSyncBackgroundTaskRequest: BGProcessingTaskRequest {
+    public static let ID = "com.instructure.icanvas.horizon-sync"
```
szabinst, line 23: let's call this `com.instructure.icanvas.career-sync`

Student/Student/Info.plist
```diff
@@ -5,6 +5,7 @@
 	<key>BGTaskSchedulerPermittedIdentifiers</key>
 	<array>
 		<string>$(PRODUCT_BUNDLE_IDENTIFIER).offline-sync</string>
+		<string>$(PRODUCT_BUNDLE_IDENTIFIER).horizon-sync</string>
```
szabinst, line 8: let's call this `career-offline-sync`

Student/Student/StudentAppDelegate.swift
```diff
@@ -185,8 +189,11 @@ class StudentAppDelegate: UIResponder, UIApplicationDelegate, AppEnvironmentDele
         CoreWebView.stopCookieKeepAlive()
         BackgroundVideoPlayer.shared.background()
         environment.refreshWidgets()
-
-        OfflineSyncScheduleInteractor().scheduleNextSync()
+        if AppEnvironment.shared.app == .horizon {
+            HSyncScheduleInteractorLive().scheduleNextSync()
+        } else {
+            OfflineSyncScheduleInteractor().scheduleNextSync()
+        }
```
vargaat, line 196: - With this if-else condition based on the currently logged in user, either academic or horizon accounts will sync but if we have mixed academic/horizon users one of them won't sync.
- If we trigger two background syncs, the OS will throttle down and one of them won't run because the background wake-up frequency would be too high.
- The academic sync logic already iterates through all accounts and syncs each one where this feature is turned on. I think the best approach would be to share the offline sync background task between the two experiences and execute the proper sync logic based on the user's experience.

vargaat (changes requested):

szabinst:
@claude

Core/Core/Features/CourseSync/BackgroundUpdates/CourseSyncBackgroundUpdatesAssembly.swift
```diff
@@ -20,9 +20,24 @@ import BackgroundTasks
 
 public enum CourseSyncBackgroundUpdatesAssembly {
 
-    public static func makeOfflineSyncBackgroundTask() -> BackgroundTask {
-        OfflineSyncBackgroundTask(syncableAccounts: OfflineSyncAccountsInteractor(),
-                                  sessions: LoginSession.sessions)
+    public static func makeOfflineSyncBackgroundTask(
+        horizonSelectedItemsInteractor: CourseSyncSelectorInteractor,
+        horizonSyncInteractor: CourseSyncInteractor
```
vargaat: I would move these to the method body where we have the academic interactors as well to avoid these being created unnecessarily for academic accounts. If you need these for testing, then convert them to a factory method.

vargaat (approved):
As we discussed, this solution won't work properly when academic and career accounts are both logged in to the app. This is because `syncNextAccount` logs in the user but the `app` itself is not updated in the `userDidLogin` method. So, if an academic account was used as the latest one, the app will start in `.student` mode and even if the sync process reaches a career account it will sync academic data for them since the `app == .horizon` condition will fail.

Please fix this and the below comment in a follow-up task.

Ahmed-Naguib93:
> As we discussed, this solution won't work properly when academic and career accounts are both logged in to the app. This is because `syncNextAccount` logs in the user but the `app` itself is not updated in the `userDidLogin` method. So, if an academic account was used as the latest one, the app will start in `.student` mode and even if the sync process reaches a career account it will sync academic data for them since the `app == .horizon` condition will fail.
> 
> Please fix this and the below comment in a follow-up task.



> As we discussed, this solution won't work properly when academic and career accounts are both logged in to the app. This is because `syncNextAccount` logs in the user but the `app` itself is not updated in the `userDidLogin` method. So, if an academic account was used as the latest one, the app will start in `.student` mode and even if the sync process reaches a career account it will sync academic data for them since the `app == .horizon` condition will fail.
> 
> Please fix this and the below comment in a follow-up task.

Jira ticket for make it later [CLXR-586](https://instructure.atlassian.net/browse/CLXR-586)

[CLXR-586]: https://instructure.atlassian.net/browse/CLXR-586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

szabinst (approved):

PR #3984, opened 2026-04-10
[MBL-19905][S] NGC - InstUI primitives

vargaat:
## PR Info

release note: none

## Test Plan
- Check the storybook in developer menu.
- Read the [README.md](https://github.com/instructure/canvas-ios/blob/a901a64c90899551553fbdcad40eebd3fb6deb90/packages/InstUI/README.md)
- You can try how name lookups work but we won't use these primitives directly.

## Screenshots

https://github.com/user-attachments/assets/dd2d185b-dbec-40af-82f8-96dc2f026a13


packages/InstUI/Sources/Primitives/InstUI.Primitives.FontWeights.Storybook.swift
```diff
@@ -0,0 +1,46 @@
...
+import SwiftUI
+
+public extension InstUI.Primitives.FontWeights {
+
+    struct Storybook: View {
+
+        public var body: some View {
+            List(InstUI.Primitives.FontWeights.all, id: \.name) { entry in
+                HStack(spacing: 12) {
+                    Text(entry.name)
```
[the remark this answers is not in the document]
vargaat: @claude `entry.name` isn't a `LocalizedStringKey` just a simple variable that won't trigger localization lookup.

scripts/instui/build-instui.js
```diff
@@ -0,0 +1,69 @@
...
+
+const INSTUI_VERSION = 'v11.7.1'
+const TOKENS_BASE_URL = `https://raw.githubusercontent.com/instructure/instructure-ui/${INSTUI_VERSION}/packages/ui-scripts/lib/build/tokensStudio`
+
+function download(url) {
+  return new Promise((resolve, reject) => {
+    https.get(url, res => {
+      let data = ''
+      res.on('data', chunk => { data += chunk })
+      res.on('end', () => resolve(data))
```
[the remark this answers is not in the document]
vargaat: Improved.

suhaibabsi-inst (dismissed):

suhaibabsi-inst:
@vargaat 
Thank you. Tremendous work 👍 
The StoryBook screens also can be very helpful !

packages/InstUI/Sources/Primitives/Generated/InstUI.Primitives.Colors.swift
```diff
@@ -0,0 +1,1014 @@
...
+    static let navy180 = Color(hex: "#061C30")
+    static let whiteOpacity10 = Color(hex: "#FFFFFF1A")
+    static let whiteOpacity20 = Color(hex: "#FFFFFF33")
+    static let whiteOpacity75 = Color(hex: "#FFFFFFBF")
+    static let greyOpacity75 = Color(hex: "#1C222BBF")
+    static let navyOpacity10 = Color(hex: "#2344651A")
+    static let greyOpacity10 = Color(hex: "#1C222B1A")
+
+    // `name` is the full dotted token path, e.g. "color.honey.honey110"
+    internal static let all: [(name: String, color: Color)] = [
```
rh12: Could we move this `all` to the end of the file into it's own extension with a MARK like `Storybook support` or something like that?
Same for the other primitives as well.

packages/InstUI/Sources/Utils/Storybook/CheckeredTexture.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import SwiftUI
+import UIKit
+
+extension Image {
+    static let checkeredTile: Image = {
```
rh12, line 23: IMHO we should move this under a Utils/Storybook folder (or Storybook/Utils or something) and also add a comment to it that it is intended for Storybook use only.

packages/InstUI/Sources/InstUI.Storybook.swift
```diff
@@ -0,0 +1,66 @@
...
+        public init() {}
+
+        public var body: some View {
+            List {
+                Section(header: Text(verbatim: "Primitives")) {
+                    NavigationLink {
+                        InstUI.Primitives.Colors.Storybook()
+                    } label: {
+                        Text(verbatim: "Colors")
+                    }
```
rh12: Maybe something like this could be created now, since we will have a lot of these storybook entries:
`StorybookItem("Colors") { InstUI.Primitives.Colors.Storybook() }`

packages/InstUI/README.md
```diff
@@ -0,0 +1,178 @@
...
+│       ├── Inclusive_Sans/
+│       └── Atkinson_Hyperlegible_Next/
+└── README.md
+```
+
+---
+
+## Public API
+
+Everything lives under the `InstUI` namespace, scoped further under `InstUI.Primitives`.
```
rh12, line 41: This last part will need to be updated later when adding each layer I guess or we could just remove it now maybe.

packages/InstUI/Sources/Primitives/Generated/InstUI.Primitives.Colors.swift
```diff
@@ -0,0 +1,1014 @@
...
+        ("color.whiteOpacity75", whiteOpacity75),
+        ("color.greyOpacity75", greyOpacity75),
+        ("color.navyOpacity10", navyOpacity10),
+        ("color.greyOpacity10", greyOpacity10)
+    ]
+}
+
+private typealias _PrimitiveColors = InstUI.Primitives.Colors
+
+public extension Color {
```
rh12: Should we have a `ShapeStyle` extension for these as well, or it makes sense only for the next layer?

packages/InstUI/Sources/InstUI.swift
```diff
@@ -0,0 +1,27 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+public enum InstUI {
+    public enum Primitives {
+        public enum Colors {}
+        public enum Sizes {}
+        public enum FontWeights {}
+        public enum Opacities {}
+        public enum FontFamilies {}
```
rh12: Do we need these separate declaration for all members here? Shouldn't we just put them in their own files where they are defined anyway? This kind of provides an overview (although the MD and the folder structure also have it), but it needs a separate maintenance and makes "Jump to definition" hard to use.

packages/InstUI/README.md
```diff
@@ -0,0 +1,178 @@
...
+A Swift Package that exposes Instructure design-system **primitives** — raw design tokens (colors, sizes, font weights, font families, opacities) — as typed Swift constants for use in SwiftUI and UIKit.
+
+Tokens are sourced from the [instructure-ui](https://github.com/instructure/instructure-ui) repository and generated automatically via `yarn build-instui`. Do not edit the generated source files by hand.
+
+---
+
+## Requirements
+
+- iOS 17+
+- Swift 6
```
rh12: Not sure it matters much, but it's basically iOS18 now and probably Swift 5

rh12 (commented):
Nicely done, thank you!

I have some comments below, will continue reviewing a bit later.

packages/InstUI/Sources/Primitives/Generated/InstUI.Primitives.Colors.swift
vargaat: This is also to help the semantic layer to look up colors, those are referenced in this format:
```
"success": {
    "value": "{color.green.green100}",
    "type": "color"
},
```

packages/InstUI/Sources/Utils/Storybook/CheckeredTexture.swift
vargaat: This is internal as all other utils and not visible from outside so it won't cause any confusion.

packages/InstUI/Sources/InstUI.Storybook.swift
vargaat: Done.

packages/InstUI/README.md
vargaat: Let's update the readme as we go through the implementation as we see fit. Currently, it gives a good overview but probably we won't need that much details especially on the primitives later on.

packages/InstUI/Sources/Primitives/Generated/InstUI.Primitives.Colors.swift
vargaat: Good idea but let's wait on usage patterns. Most probably we won't often use these primitives directly. Also worth noting, that within ShapeStyle we can't have nested enums so we can't separate those primitives from other colors (AcademicUI, HorizonUI, or later the InstUI semantic tokens like .backgroundDark).

packages/InstUI/Sources/InstUI.swift
vargaat: I created it this way to define the skeleton of the hierarchy for a quick overview but Xcode's jump to definition is definitely an often used feature that gets unusable with this so I removed them from here.

packages/InstUI/README.md
vargaat: These are unnecessary details in the readme, so I removed them. I updated the `Package.swift` to reflect these. We can't bump it to iOS 18 yet, since the main targets are on iOS 17. The Swift 6 conformance comes for free at this moment and is compatible with Swift 5, so why not?

packages/InstUI/Sources/Utils/Storybook/CheckeredTexture.swift
vargaat: Moved to a sub-folder.

petkybenedek (approved):

rh12 (approved):
QA + 1

Thanks for the typealias and the other changes!

PR #3990, opened 2026-04-15
[Chore] Change make sync and secret install steps

vargaat:
## PR Info
- During ad-hoc PR builds, secrets were installed first that copied `GoogleService-Info.plist` files into the app.
- The next step was `make provision-ci` that generates empty `GoogleService-Info.plist` files, effectively overwriting the installed secrets.
- The fix is to change the order of the two steps so those plist files won't get overwritten.

[ignore-pr-lint]


petkybenedek (approved):

rh12 (approved):
QA + 1

PR #3997, opened 2026-04-17
[MBL-19905] NGC - InstUI Semantic Layer

vargaat:
## PR Info

- Introduces `InstUI.Semantic` a runtime-loaded, role-based token layer sitting above the existing primitives. Semantic tokens (colors, sizes, spacing, border radii/widths, font sizes, opacities, font weights, font families) are loaded at app start from versioned JSON files bundled in the package, keeping values decoupled from the Swift type definitions.
- Tokens are accessed via `@Environment(\.instUITheme)` and the `InstUI.Theme` struct. Code generation (`yarn build-instui`) produces the Swift struct shapes and the bundled JSON in one pass.
- Also adds Storybook entries for all semantic token types and updates the README to document both layers and the code generation flow.

release note: none
test plan:
- Check the storybook for semantic tokens.
- Take a look at the directory structure.


packages/InstUI/Sources/Utils/Color+Inits.swift
```diff
@@ -39,4 +40,34 @@ extension Color {
...
+              let a = Double(parts[3]) else {
+            self.init(.sRGB, red: 1, green: 0, blue: 1, opacity: 1)
+            return
+        }
+        self.init(.sRGB, red: r / 255, green: g / 255, blue: b / 255, opacity: a)
+    }
+
+    /// Initializes a light/dark adaptive `Color` resolved at render time via `UITraitCollection`.
+    init(light: Color, dark: Color) {
+        self = Color(UIColor { traits in
```
petkybenedek, line 69: nit: self.init(UIColor { ... })
seems more straightforward for me

petkybenedek (dismissed):
QA +1

vargaat:
Updated the following after a conversation with @rh12:
- Removed InstUI version from token folder name. This would have made tracking changes on InstUI version update impossible. Instead I added a file that has the InstUI version in its name so its visible from Xcode.
- Removed `theme` environment variable. This solution and the `Theme.default` instance was confusing and could have get out of sync so we only kept the more versatile shared instance.
- Changed to singular form when naming variables and namespaces.

packages/InstUI/Resources/Tokens/Semantic/Layout/default.json
```diff
@@ -0,0 +1,449 @@
...
+          "type": "sizing"
+        },
+        "md": {
+          "value": "{size.size24}",
+          "type": "sizing"
+        },
+        "lg": {
+          "value": "{size.size24}",
+          "type": "sizing"
+        }
```
suhaibabsi-inst, line 40: Those seems to have the same value 🤔 
Nothing seems broken with JS scripts though.

suhaibabsi-inst (approved):
Code +1
Thanks @vargaat , that's a tremendous effort !

packages/InstUI/Resources/Tokens/Semantic/Layout/default.json
vargaat: This means that we offer the possibility for designers to create layouts with different `choiceControl` heights but this particular theme just forces all such components to be the same height across the app. If we would have a single height here, it wouldn't be possible in the InstUI system to have different `choiceControl` heights at all. The capability is here, we just don't use it yet.

packages/InstUI/Sources/TokenLoading/InstUI.TokenExtractor.swift
```diff
@@ -0,0 +1,82 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import Foundation
+
+extension InstUI {
+    public typealias TokenKey = String
```
rh12, line 22: Could we please make this non public? It should not be needed from outside.

packages/InstUI/Sources/Semantic/InstUI.Semantic.ColorLoader.swift
```diff
@@ -0,0 +1,57 @@
...
+            self.darkURL = darkURL
+        }
+
+        public func load() throws -> InstUI.Semantic.Color {
+            let lightData = try Data(contentsOf: lightURL)
+            let darkData = try Data(contentsOf: darkURL)
+            let resolver = InstUI.ColorResolver()
+            let lightLeaves = try InstUI.TokenExtractor.extractLeaves(from: lightData, section: "color")
+            let darkLeaves = try InstUI.TokenExtractor.extractLeaves(from: darkData, section: "color")
+            func token(_ path: InstUI.TokenKey) throws -> SwiftUI.Color {
```
rh12, line 48: 1. Could we move this nested method out? Basically load() could just call `InstUI.Semantic.Color.build()` with the private method.
2. Could we rename `token()` to `valueForToken()`? Also (and especially) inside the `InstUI.Semantic.Color`'s `.build()` method itself. And the other similar build methods for the other types.

packages/InstUI/Sources/InstUI+Accessors.swift
```diff
@@ -0,0 +1,79 @@
...
+    public static var iuiSemanticFontFamily: InstUISemanticFontFamily { _iuiSemanticFontFamily() }
+}
+
+extension Double {
+    public typealias iuiPrimitiveOpacity = Double.InstUI.Primitive
+    public static var iuiSemanticOpacity: InstUISemanticOpacity { _iuiSemanticOpacity() }
+}
+
+extension CGFloat {
+    public typealias iuiPrimitiveSize = CGFloat.InstUI.Primitive
```
rh12, line 47: For the generic types like these 3, IMHO it would be clearer to rename the types like this:
`CGFloat.InstUI.Primitive` -> `CGFloat.InstUI.PrimitiveSize`
Especially for the string-fontfamily.
This does not affect the aliases though.

packages/InstUI/Sources/Semantic/InstUI.Semantic.LayoutLoader.swift
```diff
@@ -0,0 +1,106 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import CoreGraphics
+import Foundation
+import SwiftUI
+
+public extension InstUI.Semantic {
```
rh12, line 23: 1. non-public extension, public members
2. same for ColorLoader

packages/InstUI/Sources/Semantic/InstUI.Semantic.LayoutLoader.swift
```diff
@@ -0,0 +1,106 @@
...
+        public let opacity: Opacity
+        public let fontWeights: FontWeight
+        public let fontFamilies: FontFamily
+    }
+
+    final class LayoutLoader: Sendable {
+        private let url: URL
+
+        public init() throws {
+            guard let url = Bundle.module.url(forResource: "default", withExtension: "json") else {
```
rh12, line 40: would it make sense to rename the json file itself from `default` to `defaultLayout` or similar?
`default` is pretty generic for a resource file which is fetched simply by name.

packages/InstUI/Sources/Semantic/InstUI.Semantic.LayoutLoader.swift
```diff
@@ -0,0 +1,106 @@
...
+                throw InstUI.TokenLoadError.missingFile("Bundled layout JSON not found")
+            }
+            self.url = url
+        }
+
+        public init(url: URL) {
+            self.url = url
+        }
+
+        public func load() throws -> LayoutSections {
```
rh12, line 50: Private methods would be IMHO much clearer here as well, same comment as in ColorLoader

packages/InstUI/Sources/Semantic/Storybook/InstUI.Semantic.Colors.Storybook.swift
```diff
@@ -0,0 +1,92 @@
...
+
+import SwiftUI
+import UIKit
+
+public extension InstUI.Semantic.Color {
+
+    struct Storybook: View {
+
+        public var body: some View {
+            List {
```
rh12, line 27: Some improvement ideas for the future:
1. search field
2. light+dark mode variants displayed at the same time (or a theme toggling button in the navbar)

packages/InstUI/Sources/Semantic/Storybook/InstUI.Semantic.FontFamilies.Storybook.swift
```diff
@@ -0,0 +1,63 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+import SwiftUI
+
+public extension InstUI.Semantic.FontFamily {
```
rh12, line 21: It just occurred to me: should we handle different line spacings?
This is just a reminder to doublecheck on that.

petkybenedek (approved):

rh12 (approved):
QA + 1

I had some comments below, but as discussed, these are just finetunings and minor stuff, they can wait for a future PR.

PR #4002, opened 2026-04-21
[MBL-19905] NGC - InstUI Component Tokens

vargaat:
## PR Info

release note: none

## Add component-level design tokens (InstUI component layer)

Adds a third token layer to InstUI, sitting above Semantic and Primitive: **Component**. Each UI component from `instructure-ui` (Badge, TextInput, Checkbox, etc.) now has a typed Swift struct collecting all its design decisions — colors, sizes, spacing, typography — resolved against the active semantic theme.

### What changed

**New layer: `InstUI.Component`**
- 65+ component token JSON files bundled as package resources (`Resources/Tokens/Component/`)
- One generated Swift file per component (`Sources/Component/Generated/`) with a typed struct and dot-syntax accessors
- `InstUI.Theme.Components` aggregates all components; accessible via `InstUI.Theme.default.components`
- `InstUI.Component.Resolver` resolves component token references through the semantic layer

**Accessor API**
- Component tokens: `.iui.<component>.<token>` (e.g. `.iui.badge.borderRadius`, `.iui.checkbox.borderColor`)
- Semantic tokens renamed for consistency: `.iuiSemantic`, `.iuiSemanticSize`, `.iuiSemanticSpacing`, etc.
- Primitive tokens: `.iuiPrimitive`, `.iuiPrimitiveSize`, `.iuiPrimitiveOpacity`, etc.

**Resolver improvements**
- `SizeResolver` now handles `em` and `px` units in addition to `rem`
- `FontWeightResolver` now accepts direct CSS numeric weights (`"400"`, `"700"`, etc.)
- Token key whitespace is trimmed before lookup

**Code generation**
- `build-instui.js` fetches component JSONs from the GitHub API at the pinned `INSTUI_VERSION`
- New `sd.config.component.js` generates all component Swift files; the `Generated/` directory is wiped and rebuilt on each run to stay in sync with upstream

### Intended use

```swift
// Component layer — preferred for implementing a specific component
RoundedRectangle(cornerRadius: .iui.badge.borderRadius)
    .fill(.iui.badge.color)
```

## Test Plan
- Since there are lot of copied / generated files in this PR, I suggest to check out the branch to take a look at the structure, and try out the various helpers.
- There's no storybook for this layer. I think it would have been too much code with small benefits. When we create each actual SwiftUI component in the future we will create storybook entries for those entities.


suhaibabsi-inst (dismissed):
Looks good to me.
And perhaps, we can think of something to limit downloading (and code generation) to only those components that are relevant to Mobile. Apparently, a lot of those are tailored to Web.

petkybenedek (dismissed):

packages/InstUI/Sources/Component/InstUI+ComponentAccessors.swift
rh12, line 1: Could we please rename to `InstUI.Component+Accessors.swift`
Same for the other layers

packages/InstUI/Sources/Component/InstUI+ComponentAccessors.swift
```diff
@@ -0,0 +1,47 @@
...
+    public enum iui {} // Namespace for the component layer
+}
+
+extension CGFloat {
+    public enum iui {} // Namespace for the component layer
+}
+
+// Called from generated component accessor extensions — module-level scope keeps `InstUI`
+// unambiguous even though each Swift type has a shadowing `Color.InstUI` / `CGFloat.InstUI` etc.
+func _iuiComponents() -> InstUI.Theme.Components { InstUI.Theme.default.components }
```
rh12, line 47: Could we please move this inside each generated file?
Mainly to avoid it being available internally for the actual components, for example.

packages/InstUI/Sources/Semantic/InstUI+SemanticAccessors.swift
```diff
@@ -68,9 +62,9 @@ public typealias InstUISemanticBorderRadius = InstUI.Semantic.BorderRadius
 public typealias InstUISemanticBorderWidth = InstUI.Semantic.BorderWidth
 public typealias InstUISemanticFontSize = InstUI.Semantic.FontSize
 
-private func _iuiSemanticColors() -> InstUISemanticColor { InstUI.Theme.default.colors }
-private func _iuiSemanticFontWeight() -> InstUISemanticFontWeight { InstUI.Theme.default.fontWeights }
-private func _iuiSemanticFontFamily() -> InstUISemanticFontFamily { InstUI.Theme.default.fontFamilies }
+private func _iuiSemanticColor() -> InstUISemanticColor { InstUI.Theme.default.color }
+private func _iuiSemanticFontWeight() -> InstUISemanticFontWeight { InstUI.Theme.default.fontWeight }
+private func _iuiSemanticFontFamily() -> InstUISemanticFontFamily { InstUI.Theme.default.fontFamily }
```
rh12, line 67: Could we remove these private helpers (each used only once), and write their content simply inline?

packages/InstUI/README.md
```diff
@@ -10,105 +10,121 @@ Tokens originate from the [instructure-ui](https://github.com/instructure/instru
...
+            .frame(width: .iui.checkbox.size, height: .iui.checkbox.size)
+    }
+}
+```
+
+Direct access through the theme is also available:
+
+```swift
+let tokens = InstUI.Theme.default.components.checkbox
+```
```
rh12, line 54: Should this have some accessor as well?

packages/InstUI/README.md
rh12, line 133: Maybe this could go to the top

packages/InstUI/Sources/Component/Generated/InstUI.Component.AiInformation.swift
```diff
@@ -0,0 +1,104 @@
...
+
+extension InstUI.Component.AiInformation {
+
+    static func load(using theme: InstUI.Theme) throws -> Self {
+        guard let url = Bundle.module.url(forResource: "AiInformation", withExtension: "json") else {
+            throw InstUI.TokenLoadError.missingFile("AiInformation.json not found in bundle")
+        }
+        let data = try Data(contentsOf: url)
+        let leaves = try InstUI.TokenExtractor.extractLeaves(from: data, section: "aiInformation")
+        let resolver = InstUI.Component.Resolver(theme: theme, leaves: leaves)
```
rh12, line 98: 1. I may be wrong, but are these used somewhere? Or these are just leftovers?
2. If used, could we extract all this to a `Resolver.init(theme: Theme, resource: String, section: String)`?

packages/InstUI/Sources/Component/Generated/InstUI.Theme.Components.swift
```diff
@@ -0,0 +1,408 @@
...
+        public let tray: InstUI.Component.Tray
+        public let treeBrowser: InstUI.Component.TreeBrowser
+        public let truncateText: InstUI.Component.TruncateText
+        public let view: InstUI.Component.View
+    }
+}
+
+extension InstUI.Theme.Components {
+
+    static func load(
```
rh12, line 98: 1. I think we should move the loading logic itself into a ComponentLoader or similar, and only leave the parts here which actually needs to be generated.
This would match the structure of the other layers, where the generated code provides only the input for these loaders.

2. I'm also thinking to extract all Theme and loading code to it's own root folder. (Even if we leave the types themselves as is.) This would make sense to me, since Theme is also a layer similarly to primivite/semantic/component layers. Also, then those layers could be solely about defining the tokens and providing accessors.

packages/InstUI/Sources/Component/Generated/InstUI.Theme.Components.swift
```diff
@@ -0,0 +1,408 @@
...
+                throw InstUI.TokenLoadError.missingFile("\(resource).json not found in bundle")
+            }
+            let data = try Data(contentsOf: url)
+            let leaves = try InstUI.TokenExtractor.extractLeaves(from: data, section: section)
+            return try build(base.withLeaves(leaves))
+        }
+
+        let aiInformation = try resolveComponent("AiInformation", "aiInformation") { resolver in
+            try InstUI.Component.AiInformation.build(dimension: resolver.dimension, color: resolver.color)
+        }
```
rh12, line 136: I'm not sure which is the final intended usage for resolving/building the Component tokens, but I was thinking to leave the actual resolving logic here (and move it out of `Generated`, see my other comment), and define the `build()` method in each generated component file, passing in only the resolver.

packages/InstUI/Sources/TokenLoading/InstUI.FontWeightResolver.swift
rh12, line 1: Loaders, resolvers, etc are not described in the Readme, which is good, since they are implementation detail.
But IMHO they should have doc-comments & inline comments to explain what they do.
Even if we happen to end up creating an implementation level documentation, IMHO it helps understanding if these things are right there in the code.

rh12 (approved):
I made some comments below to be checked (and discussed) during the consolidation work.

PR #4005, opened 2026-04-23
[MBL-19957][S] Speed up courses widget load

vargaat:
## PR Info

refs: [MBL-19957](https://instructure.atlassian.net/browse/MBL-19957)
affects: Student
release note: Courses widget loads faster on the new learner dashboard.

### Problem

The courses widget fetched unread announcement counts by calling the REST announcements endpoint, downloading every announcement for every enrolled course, and filtering client-side. This was slow.

#### GraphQL-based unread count fetch

Replaced the REST approach with a GraphQL query (`GetUnreadCourseAnnouncementCountRequest`) that asks the server directly which announcements are unread via the `participant { read }` field. Results are stored in a new CoreData entity, `CDUnreadCourseAnnouncementCount`, keyed by course ID and holding the set of unread announcement IDs (as a sorted comma-separated string for deterministic storage).

#### Pagination

The GraphQL query uses a two-phase strategy to handle per-course cursor pagination correctly:

- **Phase 1** — A single `allCourses` query (no cursor) fetches the first page for all courses simultaneously.
- **Phase 2** — For each course with `hasNextPage: true`, a targeted `courses(ids:)` query is fired with that course's own cursor. Requests within each round are fired in parallel and the process recurses until no course has a next page.

Persistence uses a smart upsert: only entities absent from the new response are deleted, so existing badge counts are never briefly zeroed during a refresh.

#### Mark-as-read on open

When a user opens an announcement, `CDUnreadCourseAnnouncementCount.removeAnnouncementId` is called alongside the existing `DiscussionTopic.markAsRead`, keeping the badge count in sync without waiting for the next background refresh.

#### Double refresh fix

Changed `LearnerDashboardScreen` from `.onAppear` to `.onNonFirstAppear` (a new `View` modifier) to avoid double refresh when the screen appears for the first time: one refresh after widgets load and another one when the screen appears.

#### Parallelization

Since we no longer depend on course IDs (that were required for the REST API announcements call) we can query announcements in parallel with the courses further reducing the overall load time.

## Test Plan

- Check if unread announcement counts match to what we've have before with the old API.
- Have a course with 20+ announcements to check paging.
- Check if courses widget actually appears faster on first load (or when the cache is expired) compared to the previous version.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-19957]: https://instructure.atlassian.net/browse/MBL-19957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


suhaibabsi-inst (commented):
Code +1

petkybenedek (dismissed):
QA +1

rh12 (approved):
QA + 1
Thanks for the paging "disabling"!

PR #4007, opened 2026-04-23
[MBL-19961][S/T/P] Pendo regions

rh12:
## PR Info

refs: [MBL-19961](https://instructure.atlassian.net/browse/MBL-19961)
release note: none

## What's new
See ticket for details.

## Test Plan
The API is not available yet, so the apikey should fall back to the current one.
These can be tested via checking the requests:
- if the user settings response is unchanged: Pendo should still work requests should be sent out
- if the user settings response has some non-empty string for the given apps apikey: No Pendo requests should be sent out

## Checklist
- [ ] Tested on phone

[MBL-19961]: https://instructure.atlassian.net/browse/MBL-19961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Common/CommonModels/Analytics/Pendo/PendoAnalyticsTracker.swift
+++ b/Core/Core/Common/CommonModels/Analytics/Pendo/PendoAnalyticsTracker.swift
@@ -36,7 +36,7 @@ public final class PendoAnalyticsTracker {
 
     private let interactor: AnalyticsMetadataInteractor
     private let pendoManager: PendoManagerWrapper
-    private let pendoApiKey: String?
+    private var pendoApiKey: String?
 
     private var isSetupCalled: Bool = false
     private var isSessionInProgress: Bool = false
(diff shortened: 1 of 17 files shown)


Core/Core/Common/CommonModels/Analytics/Pendo/PendoAnalyticsTracker.swift
```diff
@@ -57,6 +57,10 @@ public final class PendoAnalyticsTracker {
         pendoManager.initWith(url)
     }
 
+    public func storeApiKey(_ apiKey: String) {
+        pendoApiKey = apiKey.nilIfEmpty
```
suhaibabsi-inst, line 61: Perhaps, we need to keep old value in case retrieved api key was empty string ?

```suggestion
        pendoApiKey = apiKey.nilIfEmpty ?? pendoApiKey
```

rh12: The key is validated at the callsite (it won't be empty), I just left this in place here to match the init method.
This will be updated a bit when the backend releases the actual change which provides the key.
Then the key will be removed from the init altogether. This current logic is required to handle the temporary case that the backend is not updated but the new app version is out already.

vargaat (approved):
QA+1, Fallback works

petkybenedek (approved):
Code +1

PR #4012, opened 2026-04-27
[MBL-19974][S] Offline Mode - Continue Studio iframe tags replacement when some videos are not downloaded

suhaibabsi-inst:
## PR Info

refs: [MBL-19974](https://instructure.atlassian.net/browse/MBL-19974)
builds: Student
affects: Student
release note: None.

## Changes

- The iframe replacement loop previously threw `offlineVideoIDNotFound` as soon as it encountered any iframe with no matching offline video, discarding all replacements computed so far and never writing the file.
- This fix collects unmatched iframes into a `nonFoundFrames` array, continues replacing the rest, writes the updated HTML to disk, and only then throws `offlineVideoIDNotFound` if any frames were unmatched.
- Net result: pages with a mix of matched and unmatched iframes now correctly render the videos that did download offline.

## Test Plan

- [ ] Sync a course offline that has a page with **two or more Studio videos**. See ticket's description for test account.
- [ ] Confirm all synced Studio videos play inline as `<video>` elements when offline.
- [ ] Simulate the failure case: if one video is missing from the offline sync, verify the remaining videos on the same page still render correctly.

## Checklist

- [ ] A11y checked
- [ ] Tested on phone
- [ ] Tested on tablet
- [ ] Tested in dark mode
- [ ] Tested in light mode
- [ ] Approve from product

[MBL-19974]: https://instructure.atlassian.net/browse/MBL-19974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Features/Studio/Model/StudioIFrameReplaceInteractor.swift
+++ b/Core/Core/Features/Studio/Model/StudioIFrameReplaceInteractor.swift
@@ -57,9 +57,11 @@ public class StudioIFrameReplaceInteractorLive: StudioIFrameReplaceInteractor {
             throw StudioIFrameReplaceError.failedToConvertDataToString
         }
 
+        var nonFoundFrames: [StudioIFrame] = []
         for iframe in iframes {
             guard let offlineVideo = offlineVideos.first(where: { $0.ltiLaunchID == iframe.mediaLTILaunchID }) else {
-                throw StudioIFrameReplaceError.offlineVideoIDNotFound
+                nonFoundFrames.append(iframe)
+                continue
             }
             htmlString = replaceStudioIFrame(
                 html: htmlString,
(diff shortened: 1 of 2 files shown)


rh12 (commented):
Code + 1

rh12 (approved):
QA + 1

Core/Core/Features/Studio/Model/StudioIFrameReplaceInteractor.swift
```diff
@@ -77,6 +79,10 @@ public class StudioIFrameReplaceInteractorLive: StudioIFrameReplaceInteractor {
         } catch {
             throw StudioIFrameReplaceError.failedToSaveUpdatedHtml
         }
+
+        if nonFoundFrames.isNotEmpty {
```
vargaat, line 83: A `bool` would have been enough instead of the array.

vargaat (approved):
QA+1

PR #4016, opened 2026-04-28
[MBL-19904] NextGen - Preparations

vargaat:
## PR Info

refs: [MBL-19904](https://instructure.atlassian.net/browse/MBL-19904)
affects: Student
release note: none

## What's new?
- Introduced NextGen, a new iOS framework flavor of the Student app (alongside Horizon), with its own XcodeGen project, SwiftLint config, unit test target, test plan, and workspace integration.
- Added .nextgen as a new AppEnvironment.App case, wired up across all relevant switch statements to match Student behavior.
- Added a nextgen_student feature flag that sets AppEnvironment.app = .nextgen at login, which limits the Student tab bar to the dashboard only and hides the tab bar chrome. This is because the navigation for the new gen experience is not yet been decided, so it's just there as a visible placeholder change.
- Introduced NextGenRoutes — a routing layer that merges NextGen-specific route overrides with the academic routes, deduplicating any overridden templates so NextGen handlers always win.
- Exposed Router.handlers as public let to allow route composition without duplicating the route array.
- Hidden the "opt out of new dashboard" toggle in LearnerDashboardSettingsScreen when running as NextGen. This was a requested change and will stay like this.

[MBL-19904]: https://instructure.atlassian.net/browse/MBL-19904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
+++ b/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
@@ -30,7 +30,7 @@ public protocol AppEnvironmentDelegate {
 
 open class AppEnvironment {
     public enum App: String {
-        case parent, student, teacher, horizon
+        case parent, student, teacher, horizon, nextgen
     }
 
     public internal(set) lazy var uploadManager = UploadManager(
--- a/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
+++ b/Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
@@ -30,7 +30,11 @@ public protocol AppEnvironmentDelegate {
 
 open class AppEnvironment {
     public enum App: String {
-        case parent, student, teacher, horizon, nextgen
+        case parent
+        case teacher
+        case student
+        case horizon // Student app experience
+        case nextgen // Student app experience
     }
 
     public internal(set) lazy var uploadManager = UploadManager(
--- a/Core/Core/Features/Login/GetSSOLogin.swift
+++ b/Core/Core/Features/Login/GetSSOLogin.swift
@@ -38,14 +38,12 @@ public class GetSSOLogin {
 
     static func codeForApp(_ app: App) -> String {
         switch app {
-        case .student, .nextgen:
+        case .student, .horizon, .nextgen:
             return "code"
         case .teacher:
             return "code_ios_teacher"
         case .parent:
             return "code" // "code_ios_parent"
-        case .horizon:
-            return "code"
         }
     }
 
--- a/Core/Core/Features/Login/GetSSOLogin.swift
+++ b/Core/Core/Features/Login/GetSSOLogin.swift
@@ -38,7 +38,7 @@ public class GetSSOLogin {
 
     static func codeForApp(_ app: App) -> String {
         switch app {
-        case .student:
+        case .student, .nextgen:
             return "code"
         case .teacher:
             return "code_ios_teacher"
--- a/NextGen/NextGen/nextgen.yml
+++ b/NextGen/NextGen/nextgen.yml
@@ -1,0 +1,29 @@
+---
+targets:
+  NextGen:
+    type: framework
+    platform: iOS
+    transitivelyLinkDependencies: false
+    settings:
+      APPLICATION_EXTENSION_API_ONLY: false
+      SWIFT_EMIT_LOC_STRINGS: true
+      PRODUCT_BUNDLE_IDENTIFIER: com.instructure.nextgen
+      TARGETED_DEVICE_FAMILY: 1,2
+    sources:
+      - path: ./
+        excludes:
+          - ".swiftlint.yml"
+          - "nextgen.yml"
+    dependencies:
+      - target: Core/Core
+      - package: BusinessLogic
+      - package: InstUI
+    preBuildScripts:
+      - path: ../../scripts/xcode-build-phases/swiftLint.sh
+        name: SwiftLint
+        basedOnDependencyAnalysis: false
+schemes:
+  NextGen:
+    build:
+      targets:
+        NextGen: all
--- a/NextGen/NextGen/nextgen.yml
+++ b/NextGen/NextGen/nextgen.yml
@@ -27,3 +27,7 @@ schemes:
     build:
       targets:
         NextGen: all
+    test:
+      testPlans:
+        - path: ../NextGenTests.xctestplan
+          defaultPlan: true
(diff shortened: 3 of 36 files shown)


Core/Core/Features/Login/GetSSOLogin.swift
```diff
@@ -38,7 +38,7 @@ public class GetSSOLogin {
 
     static func codeForApp(_ app: App) -> String {
         switch app {
-        case .student:
+        case .student, .nextgen:
```
rh12: Could we move `.horizon` here as well?

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
```diff
@@ -30,7 +30,7 @@ public protocol AppEnvironmentDelegate {
 
 open class AppEnvironment {
     public enum App: String {
-        case parent, student, teacher, horizon
+        case parent, student, teacher, horizon, nextgen
```
rh12: 1. IMHO at this point we could break them to their own line, and add trailing comments for the two experiences, that these are student app variations/experiences
2. Or would this be a good time to somehow separate the app itself and the experiences properly?

NextGen/NextGen/nextgen.yml
```diff
@@ -0,0 +1,33 @@
+---
+targets:
+  NextGen:
+    type: framework
+    platform: iOS
+    transitivelyLinkDependencies: false
+    settings:
+      APPLICATION_EXTENSION_API_ONLY: false
+      SWIFT_EMIT_LOC_STRINGS: true
+      PRODUCT_BUNDLE_IDENTIFIER: com.instructure.nextgen
```
rh12: Should this have it's own bundle?

NextGen/NextGenUnitTests/Helpers/NextGenTestCase.swift
```diff
@@ -0,0 +1,62 @@
...
+    var databaseClient: NSManagedObjectContext { database.viewContext }
+
+    var api: API { environment.api }
+    var environment: TestEnvironment!
+    var router = TestRouter()
+    var logger = TestLogger()
+
+    let window = UIWindow()
+
+    override func setUp() {
```
rh12, line 36: Should we add userdefaults handling here (including reset)?
Should we extract (most of?) CoreTestCase to some baseclass in TestsFoundation?

Student/Student/LearnerDashboard/Settings/ViewModel/LearnerDashboardSettingsViewModel.swift
```diff
@@ -26,6 +26,7 @@ import UIKit
 @Observable
 final class LearnerDashboardSettingsViewModel {
     var useNewLearnerDashboard: Bool
+    var isOptOutAllowed: Bool { environment.app != .nextgen }
```
rh12, line 29: Maybe this should be a switch (without default branch)

Student/Student/StudentAppDelegate.swift
```diff
@@ -430,8 +431,10 @@ extension StudentAppDelegate {
     private func setTabBarControllerFor(experience: Experience, isStartup: Bool, session: LoginSession?) {
         switch experience {
         case .academic:
-            AppEnvironment.shared.app = .student
-            AppEnvironment.shared.router = academicRouter
+            AppEnvironment.shared.app = ExperimentalFeature.nextgenStudent.isEnabled ? .nextgen : .student
+            AppEnvironment.shared.router = AppEnvironment.shared.app == .nextgen
+                ? NextGenRoutes.makeRouter(academicRoutes: academicRouter.handlers)
+                : academicRouter
```
rh12: Would it make sense to move this logic to an `AppEnvironment.setApp(to: )` method, disallow writing `app` from the outside and update all callsites (like the lazy var at the top here, and horizon below)

In that case make sure TestEnvironment still works

Student/Student/StudentTabBarController.swift
```diff
@@ -59,6 +59,11 @@ class StudentTabBarController: UITabBarController, SnackBarProvider {
         reportScreenView(for: selectedIndex, viewController: viewControllers![selectedIndex])
         addSnackBar()
         registerForTraitChanges()
+
+        if AppEnvironment.shared.app == .nextgen {
+            viewControllers = Array(viewControllers?.prefix(1) ?? [])
```
rh12, line 67: Could we please add a comment for what this does and perhaps why?

Core/Core/Features/Login/GetSSOLogin.swift
vargaat: Done.

Core/Core/Common/CommonModels/AppEnvironment/AppEnvironment.swift
vargaat: Done #1. We can think about #2 but won't do it right now.

Student/Student/StudentTabBarController.swift
vargaat: Done.

Student/Student/LearnerDashboard/Settings/ViewModel/LearnerDashboardSettingsViewModel.swift
vargaat: The business request we introduced here is that we shouldn't show the opt out switch for the nextgen experience and I think this condition clearly expresses this intention. For other types the behavior is undefined as of yet, so I'm not sure that adding explicit true/false for other apps would be better.

Student/Student/StudentAppDelegate.swift
vargaat: We can't move this logic to `AppEnvironment`, because that's in `Core` while the custom routers are in projects depending on `Core`. We could do a `set(app:router:)` method to make the `app` setter private but a lot of tests are depending on directly setting the `app` to test different behaviors. Maybe a router factory based on `App` that could be injected into `Core` would be the best solution with some proper DI container. I see the improvement point but I'm not sure it's worth the shot now.

Student/Student/LearnerDashboard/Settings/ViewModel/LearnerDashboardSettingsViewModel.swift
rh12: The point would be to make the compiler force us to choose, whenever we add a new case.
But it's not that important.

suhaibabsi-inst (dismissed):
Thank you @vargaat !
Looks good to me.

NextGen/NextGenUnitTests/Helpers/NextGenTestCase.swift
vargaat: We can use the one introduced lately to `TestEnvironment`.  Some consolidation related to these base test cases would be nice but not as part of this PR.

petkybenedek (approved):

PR #4028, opened 2026-07-24
Fix Command Injection Vulnerability in server.js

GJQ-Q:
# Command Injection Reproduction Notes for `server.js`

## Summary

Hello,

I am writing to report a potential OS Command Injection vulnerability in the following file:

`server.js`

The issue appears when user-controlled input is used to construct shell command strings that are later executed via `child_process.execSync(...)` or `child_process.exec(...)`.

The vulnerable code implements a local web server that accepts command strings via a POST request to the `/terminal` endpoint and executes them directly.

The execution flow is as follows:

```
HTTP POST request to /terminal
        |
        v
req.body.command
        |
        v
exec(command, async)
        |
        v
process.execSync(command)
```

Since the `command` variable is taken directly from the request body without any validation or sanitization, it allows arbitrary OS command execution.

## Root Cause

The issue is caused by the insecure design of the `/terminal` route, which treats user input as a shell command:

```js
app.post("/terminal", (req, res) => {
    const output = exec(req.body.command, req.query.async).toString("utf8").trim();
    res.send(output);
});

function exec(command, async) {
    if (async === "true") {
        return process.exec(command);
    } else {
        return process.execSync(command);
    }
}
```

The value from:

```js
req.body.command
```

is directly passed into:

```js
process.exec(command)
```

or:

```js
process.execSync(command)
```

without any validation or sanitization.

This allows an attacker to execute arbitrary system commands by providing a malicious command string in the POST request body.

## Reproduction Material

A reproduction script is provided in: `poc_server.js`

`poc_server.js`

This Proof of Concept (PoC) automates the process of starting the vulnerable server and sending a payload to demonstrate the vulnerability.

### What the PoC Does

The PoC performs the following steps:

1. **Starts the Target Server:**  
   It uses `child_process.spawn()` to run:

```text
node server.js
```

as a background process.

2. **Monitors Readiness:**  
   It listens to the server's stdout for the:

```text
Terminal server started.
```

message to ensure the service is active.

3. **Sends the Exploit Payload:**  
   It constructs a JSON body containing a malicious command:

```text
start calc.exe
```

4. **Triggers Execution:**  
   It sends a POST request to:

```text
http://127.0.0.1:4567/terminal
```

5. **Captures Output:**  
   It logs the server's response, confirming that the command was executed on the host system.

## Example Payload

The PoC uses the following payload:

```json
{
  "command": "start calc.exe"
}
```

In a Windows environment, this command launches the Calculator application as a visible indicator that the `req.body.command` value directly influences system-level command execution.

## How to Run

Run from the project root:

```bash
node poc_server.js
```

## Expected Output

When the PoC runs, you should see output similar to:

```text
[*] Starting target server: server.js...
[Server Log]: Terminal server started.
[!] Server is ready. Attempting to inject command...
-----------------------------------
[+] Exploit Successful! Server returned:
(Command executed successfully)
-----------------------------------
[*] Closing session...
```

In a Windows environment, the Calculator application (`calc.exe`) will be launched, demonstrating that the `req.body.command` value directly influences system-level command execution.

This shows that external input can reach the OS command execution sink without proper sanitization.

## Patch Explanation

This branch also includes a security-hardened version of `server.js` intended to mitigate the command injection risk described above.

### What the patch changes

The patch implements a robust Input Validation layer before any command execution takes place.

The fix introduces a "Security Gatekeeper" at the `/terminal` endpoint.

In the vulnerable version:

```js
req.body.command
```

was passed directly to:

```js
process.exec()
```

or:

```js
process.execSync()
```

The updated version validates the command string against a strict set of rules before execution.

The dangerous input never reaches the command execution functions.

### Validation introduced by the patch

The validation is designed to prevent command chaining and shell injection.

#### Type & Content Check

The system ensures:

- The input is a valid string.
- The input is not empty.

#### Malicious Character Filtering

The system blocks shell metacharacters commonly used for injection, such as:

```text
;
&
|
$
(
)
`
>
<
```

These characters are rejected to prevent command chaining, variable expansion, subshell execution, and redirection attacks.

#### Allowlist Pattern

The validation is tuned to only permit recognized administrative command patterns, ensuring that unauthorized binaries cannot be invoked.

If validation fails, the command is rejected instead of being passed into command execution.

### DIFF
--- a/scripts/terminal-server/server.js
+++ b/scripts/terminal-server/server.js
@@ -27,6 +27,8 @@ const port = 4567;
 
 app.use(express.json());
 
+const SAFE_INPUT_PATTERN = /^[a-zA-Z0-9\s\.\/\-\_\/\"\'\=\:\@]+$/;
+
 app.listen(port, function(err) {
     if (err) {
         console.log("Error starting terminal server.");
@@ -36,8 +38,17 @@ app.listen(port, function(err) {
 });
 
 app.post("/terminal", (req, res) => {
-    const output = exec(req.body.command, req.query.async).toString("utf8").trim();
-    res.send(output);
+const command = req.body.command;
+    if (typeof command !== 'string' || !SAFE_INPUT_PATTERN.test(command)) {
+        return res.status(403).send("Error: Forbidden characters in command.");
+    }
+
+    try {
+        const output = exec(command, req.query.async).toString("utf8").trim();
+        res.send(output);
+    } catch (err) {
+        res.status(500).send(err.message);
+    }
 });
 
 function exec(command, async) {
--- a/scripts/terminal-server/server.js
+++ b/scripts/terminal-server/server.js
@@ -20,14 +20,57 @@
 // on the /terminal path and executes the body of the request
 // as a command line statement.
 
-const process = require("child_process");
+const { spawn, spawnSync } = require("child_process");
 const express = require("express");
 const app = express();
 const port = 4567;
 
 app.use(express.json());
+const NETWORK_SERVICE_NAME_PATTERN = /^[^\0\r\n']{1,128}$/;
 
-const SAFE_INPUT_PATTERN = /^[a-zA-Z0-9\s\.\/\-\_\/\"\'\=\:\@]+$/;
+function parseAllowedCommand(command) {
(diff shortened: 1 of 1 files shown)


vargaat:
Thanks, we already fixed this on our side.

